PR #7, opened 2026-02-26
feat: user report system (NIP-56)

rabble-claw:
Adds /api/v1/report endpoint with D1 storage and auto-escalation at 3/5 unique reporter thresholds.

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -12,6 +12,7 @@ import { verifyZeroTrustJWT } from './admin/zerotrust.mjs';
 import { fetchNostrEventBySha256, parseVideoEventMetadata } from './nostr/relay-client.mjs';
 import dashboardHTML from './admin/dashboard.html';
 import swipeReviewHTML from './admin/swipe-review.html';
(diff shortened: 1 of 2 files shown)


rabble:
shouldn't reports be 1984 events over nostr?!

rabble-claw:
You're right — the primary path should be NIP-56 kind 1984 events over Nostr, not a REST API.

Funnelcake already handles this: migration 000070 has `report_counts` table + materialized view that auto-extracts reports from kind 1984 events into ClickHouse. So the Nostr-native path already exists in the relay.

The REST `/api/v1/report` endpoint is useful as:
- Admin/internal tool for manual flagging
- Bridge for platforms that don't speak Nostr
- Backfill from external report sources

But the main user-facing flow should be: client submits kind 1984 event → relay stores it → MV populates report_counts → threshold triggers moderation review.

Should I add a funnelcake hook that calls the moderation service when report thresholds are hit? Or keep it purely ClickHouse-side with periodic queries?

PR #56, opened 2026-03-29
fix: keep original vine imports serveable

rabble:
## Summary
- keep original/classic Vine imports serveable while preserving raw AI scores for training and debugging
- split downstream moderation signals from enforcement so original Vine imports can still emit nudity, violence, gore, and other non-AI moderation context
- add docs for the policy/design change and stabilize Vitest on this branch with the existing single-worker workaround from the known worker-pool flake

## Test Plan
- [x] `npx vitest run`

### DIFF
--- a/src/admin/dashboard.html
+++ b/src/admin/dashboard.html
@@ -1548,6 +1548,19 @@
   </style>
 </head>
 <body>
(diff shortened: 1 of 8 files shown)


NotThatKindOfDrLiz:
Closing — this was superseded by #63, which landed the vine-serveable fix along with the rollback and downstream signal splitting.

PR #67, opened 2026-03-31
fix: normalize Blossom failure responses across moderation endpoints

### LINKED ISSUE #65: Normalize Blossom failure responses across moderation endpoints
The three moderation endpoints now correctly return `502` when Blossom does not confirm enforcement, but their failure payloads still differ slightly and the response construction is duplicated.

Scope:
- Align `502` response fields across:
  - `/admin/api/moderate/:sha256`
  - `/api/v1/moderate`
  - `/api/v1/quarantine/:sha256`
- Extract a shared helper for Blossom failure responses to reduce drift

This is cleanup only. The production enforcement bug is already fixed in PR #57.

mbradley:
## Summary
- Extract `blossomFailureResponse()` helper used by all three moderation endpoints
- Normalizes the 502 response shape: `{ success, sha256, action, blossom_notified, error }`
- Removes inconsistencies: `previousAction` and `updated_at` dropped from error responses (callers check `success` and `blossom_notified`)

Closes #65

## Test plan
- [x] 562 tests passing
- [x] Existing 502 tests updated to match normalized shape

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -1456,17 +1456,7 @@ export default {
 
       if (!blossomResult.success && !blossomResult.skipped) {
         console.warn(`[ADMIN] Blossom notification failed: ${blossomResult.error}`);
-        return new Response(JSON.stringify({
-          success: false,
-          sha256,
-          action,
-          previousAction,
-          blossom_notified: false,
-          error: `Moderation recorded but media server did not confirm: ${blossomResult.error}`,
-        }), {
-          status: 502,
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz (approved):
LGTM.

The helper extraction and response normalization are the right cleanup here. The only follow-up I’d make is non-blocking: move `blossomFailureResponse()` so the `notifyBlossom()` JSDoc sits directly above the function it documents again.

PR #72, opened 2026-04-01
fix: restore code clobbered by stale branch merge

mbradley:
## Summary

- Merge `1bdb009` resolved conflicts using the stale `quick-review-ux-improvements` branch, which silently dropped features that had been merged to main between its fork point and the merge
- This PR restores all clobbered code and preserves the legitimate post-merge additions (ATProto labeler, CORS on public status, transcript proxy, inbound-labels stubs)
- **Most code restored here was previously approved and merged via PRs #47, #48, #49, #52, #53, #54, #57, #61, #63, #64, #67, #69**

**Restored:**
- `pendingFlagged` + `pending` breakdown in `/admin/api/stats` (dashboard "AI Flagged" count was showing 0)
- `/api/v1/notify` endpoint (DM-only notifications used by relay-manager)
- `notifyReporters` calls in admin moderate, handleModerationResult, and cron paths
- `blossomFailureResponse` helper + 502 returns on Blossom notification failures
- Classic vine rollback endpoint
- Metadata fields (`title`, `author`, `event_id`, `content_url`, `published_at`) in D1 UPSERT
- Enriched `sendModerationDM` options (categories object, title, publishedAt)
- `Cache-Control: no-store` on admin JSON responses
- Sort parameter (`oldest`/`newest`) on untriaged list
- `buildDownstreamPublishContext` import
- ~791 lines of tests covering the above

**New improvements beyond restoration:**
- `blossomFailureResponse` shared helper consolidating three inline 502 responses (PR #57 review)
- Blossom failure check moved before relay delete in admin handler to prevent partial enforcement (PR #57 review)
- 502 tests added for all three moderate endpoints (PR #57 review)
- Already-SAFE skip in classic vine rollback to avoid unnecessary Blossom webhooks (PR #63 review)
- Blossom failure handling in classic vine rollback (PR #63 review)
- CORS wrapping on `/check-result/` responses
- `deleteRelayEventIds` refactored with per-event error tracking
- `deleteEventFromRelayBySha256` now fetches all d-tag versions
- Admin delete-event route returns 502 on relay failure

## Test plan

- [x] 584/584 tests passing locally (34 test files)
- [x] Lint clean (85 modules + 4 HTML files)
- [x] Verified restored index.mjs differs from last known-good commit (97d9b63) only by the CORS fix (4 lines)
- [x] Verified index.test.mjs matches 97d9b63 exactly
- [x] CI passes
- [x] Deploy to production and verify dashboard "AI Flagged" count is non-zero

PR #75, opened 2026-04-02
fix: add ETag cache validation to admin HTML pages

mbradley:
## Summary

- Admin HTML pages (dashboard, swipe review, messages) were served without cache headers, so browsers could run stale JavaScript from cached or still-open tabs after a deploy
- API responses already had `Cache-Control: no-store` via `JSON_HEADERS`, but the HTML pages did not
- Adds ETag-based validation: `Cache-Control: no-cache` + `ETag` computed from HTML content at module load
- Browsers cache locally but revalidate every request. 304 if unchanged (saves bandwidth), full response after deploy
- ETags update automatically on deploy -- no manual version bumping needed

## Context

Through direct testing and inspection of persisted data, we confirmed that moderation decisions are saving correctly after recent fixes (PRs #72, #73). We isolated a remaining gap: browsers serving stale frontend code from cached or open tabs would not pick up deploys that included fixes and improvements, making it appear that decisions weren't persisting even after the underlying issue was resolved.

## Test plan

- [x] 588/588 tests pass
- [x] Verified only three `new Response(html, ...)` calls changed
- [x] No changes to API endpoints, D1 writes, or any other functionality

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -52,6 +52,32 @@ const ADMIN_HOSTNAME = 'moderation.admin.divine.video';
 const API_HOSTNAME = 'moderation-api.divine.video';
 const DEFAULT_RELAY_ADMIN_URL = 'https://relay.admin.divine.video';
 const JSON_HEADERS = { 'Content-Type': 'application/json', 'Cache-Control': 'no-store' };
+
+// ETags for admin HTML pages — computed once at module load, change only on deploy.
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz (approved):
LGTM.

Two follow on suggestions:
- If-None-Match handling is strict equality only (could be more RFC-tolerant for weak/multi-value tags).
- 32-bit hash collision risk is low but non-zero; stronger digest could further harden behavior.

PR #76, opened 2026-04-03
fix: set reviewed_by on /api/v1/moderate so relay-manager bans don't pollute the AI flagged queue

mbradley:
## Summary

When a moderator bans content in relay-manager, that decision should be final — it should not reappear in Aleysha's review queue, and AI should not re-process it. Both depend on \`reviewed_by\` being set in moderation-service's D1.

- \`/api/v1/moderate\` was writing the D1 record but leaving \`reviewed_by = NULL\`, so relay-manager bans were enforced in Blossom but kept surfacing in the human review queue as if no one had looked at them.
- The review queue filter is \`reviewed_by IS NULL\`. Set it, the item is gone. The AI pipeline dedup check also skips any sha256 with an existing D1 record.
- Root cause surfaced when Matt's March 24 routing fix (\`fa1f6ec\`) made relay-manager bans actually reach moderation-service for the first time. Before that they were silently 404ing.

## Changes

- \`INSERT\` and \`ON CONFLICT\` in \`/api/v1/moderate\` now set \`reviewed_by = source\` and \`reviewed_at\`
- Added test asserting \`reviewed_by\` is written with the caller's source value

## Backfill

59 existing relay-manager entries updated directly in production D1:
\`\`\`sql
UPDATE moderation_results
SET reviewed_by = 'relay-manager', reviewed_at = moderated_at
WHERE provider = 'relay-manager' AND reviewed_by IS NULL
\`\`\`

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -2881,24 +2881,31 @@ async function runMigration() {
           });
         }
 
-        // Update or insert moderation result
+        // Update or insert moderation result.
+        // reviewed_by is set to the source because /api/v1/moderate is called by human
+        // operators (relay-manager UI, external tools) — not by the AI pipeline.
+        // Without this, relay-manager bans land in the AI Flagged human-review queue.
+        const reviewedBy = source || 'external-api';
+        const now = new Date().toISOString();
         await env.BLOSSOM_DB.prepare(`
-          INSERT INTO moderation_results (sha256, action, provider, scores, categories, moderated_at)
-          VALUES (?, ?, ?, ?, ?, ?)
+          INSERT INTO moderation_results (sha256, action, provider, scores, categories, moderated_at, reviewed_by, reviewed_at)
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz:
This looks good. The `/api/v1/moderate` path now marks operator-originated actions as reviewed at write time, which fixes the queue pollution without widening the change beyond the affected UPSERT path.

The SQL update is consistent on both insert and conflict, and the backfill described in the PR is the right way to clean up existing rows.

Approved.

NotThatKindOfDrLiz:
Non-blocking: the new SQL-binding test is weaker than it looks because `provider` was already bound to the same `source` value before this change. `expect(capturedBindings).toContain("relay-manager")` will still pass even if `reviewed_by` is omitted or mis-bound.

If you touch this area again, it would be worth tightening the assertion to check the exact binding positions or total binding shape instead of value presence alone.

NotThatKindOfDrLiz (approved):

PR #77, opened 2026-04-03
fix: forward Range header in admin video proxy for age-restricted content

mbradley:
## Summary

Age-restricted videos were showing black in the moderation dashboard.

**Root cause:** PERMANENT_BAN content plays fine via Tier 1 (CDN still has cached content from when it was Active). AGE_RESTRICTED content falls through to Tier 3 (admin bypass via `BLOSSOM_WEBHOOK_SECRET`), which was fetching from Blossom without forwarding the browser's `Range` header. The browser sends `Range: bytes=0-` to start playback and expects a `206 Partial Content` response — receiving a full `200 OK` instead causes the video player to black out.

## Changes

Single change in the Tier 3 fetch in `/admin/video/` proxy (`src/index.mjs`):

- Forward `Range` header from the incoming browser request to Blossom
- Pass back `Content-Range`, `Accept-Ranges`, and `Content-Length` from Blossom's response
- Preserve response status (`206` vs `200`) so the browser video player handles it correctly

No HTML changes, no new endpoints. Matches the pattern already used in divine-relay-manager's media proxy.

## Testing

592 tests passing. Needs manual verification against a live age-restricted video in the dashboard.

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -2135,18 +2135,26 @@ export default {
         // Fall back to admin bypass endpoint which serves regardless of moderation status
         if (env.BLOSSOM_WEBHOOK_SECRET) {
           console.log(`[ADMIN] CDN returned ${cdnResponse.status}, trying admin bypass for ${sha256}`);
-          const bypassResponse = await fetch(adminBypassUrl, {
-            headers: { 'Authorization': `Bearer ${env.BLOSSOM_WEBHOOK_SECRET}` }
-          });
-          if (bypassResponse.ok) {
+          const bypassHeaders = { 'Authorization': `Bearer ${env.BLOSSOM_WEBHOOK_SECRET}` };
+          const rangeHeader = request.headers.get('Range');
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz:
This looks good. The fix stays tightly scoped to the Tier 3 admin bypass path, forwards the browser `Range` header through to Blossom, and now preserves the partial-content response metadata the video player expects for age-restricted playback.

That matches the stated root cause and avoids widening the proxy behavior beyond the affected path.

Approved.

NotThatKindOfDrLiz:
Non-blocking: `bypassResponse.status === 206` is redundant here because `Response.ok` already covers the full `200-299` range, including `206`. It is harmless, but if this code gets touched again you could simplify that branch back to `if (bypassResponse.ok)`.

NotThatKindOfDrLiz (approved):
This looks good. The fix stays tightly scoped to the Tier 3 admin bypass path, forwards the browser `Range` header through to Blossom, and now preserves the partial-content response metadata the video player expects for age-restricted playback.

Approved.

PR #85, opened 2026-04-12
Use FunnelCake REST for admin metadata lookups

rabble:
## Summary
- switch admin metadata hydration from relay WebSocket lookups to FunnelCake REST `GET /api/videos/{id}`
- prefer stored `moderation_results` metadata for already-moderated content so admin review still has context when relay lookups miss
- add regression tests covering `/admin/api/video/:id` and `/admin/api/nostr-context/:sha256` so these paths fail if they try to use WebSocket again

## Test Plan
- [x] `npm test`
- [x] `npm test -- src/uploader-enforcement.test.mjs`
- [x] `npm test -- src/index.test.mjs`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ old_files/
 .test-nsec
 old_files/
 cost-projections-internal.md
+.worktrees/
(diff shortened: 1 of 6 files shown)


rabble:
Superseded by #86, which is based cleanly on current main and contains only the admin REST metadata lookup work.

PR #90, opened 2026-04-15
fix: preserve admin review video playback ranges

rabble:
## Summary
- forward byte-range request headers through the admin video proxy
- preserve upstream partial-content response headers and status codes for CDN, transcode, and admin-bypass playback
- add regression coverage for browser playback against 

## Test Plan
- [x] npm test -- --exclude '.worktrees/**' --exclude '.claude/**' src/index.test.mjs -t "admin video proxy format fallback"
- [x] npm test -- --exclude '.worktrees/**' --exclude '.claude/**' src/index.test.mjs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ old_files/
 .test-nsec
 old_files/
 cost-projections-internal.md
+.worktrees/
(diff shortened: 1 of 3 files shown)


rabble:
Closing this because it was opened from a stale local main. Rebasing against origin/main shows the admin video Range/206 playback fix is already on main via #88 (commit a1d0ffb), and that implementation is more complete than this branch. There is no remaining delta to merge.

PR #92, opened 2026-04-16
feat: per-video delete end-to-end enforcement

mbradley:
## Summary

Moderation-service side of per-video delete end-to-end enforcement.

**Issue:** divinevideo/divine-mobile#3102
**Rollout comment:** https://github.com/divinevideo/divine-mobile/issues/3102#issuecomment-4264760572
**Rollout plan:** [`support-trust-safety/docs/rollout/2026-04-16-creator-delete-rollout.md`](https://github.com/divinevideo/support-trust-safety/blob/working-draft/docs/rollout/2026-04-16-creator-delete-rollout.md)
**Sibling PR:** divinevideo/divine-blossom#85 (Blossom DELETE action)

## What's in this PR

- D1 migration for the `creator_deletions` audit table
- NIP-98 Authorization header validator (`src/creator-delete/nip98.mjs`)
- Race-safe D1 helpers with INSERT ON CONFLICT DO NOTHING idempotency (`src/creator-delete/d1.mjs`)
- `processKind5` core function with multi-target support + state taxonomy (`src/creator-delete/process.mjs`)
- KV-backed rate limiter (`src/creator-delete/rate-limit.mjs`)
- `POST /api/delete/{kind5_id}` sync endpoint with NIP-98 auth, IP + pubkey rate limiting, 8s budget timeout (`src/creator-delete/sync-endpoint.mjs`)
- `GET /api/delete-status/{kind5_id}` status endpoint (`src/creator-delete/status-endpoint.mjs`)
- Funnelcake fetch-with-retry via REST (`src/creator-delete/funnelcake-fetch.mjs`)
- `notifyBlossom` extraction to shared `src/blossom-client.mjs` with DELETE action added
- 1-minute cron for kind 5 polling + transient retry with exponential backoff (`src/creator-delete/cron.mjs`)
- Route wiring + `CREATOR_DELETE_PIPELINE_ENABLED` feature flag (default off) in `src/index.mjs` + `wrangler.toml`
- Structured observability logs across the pipeline
- `fetchNostrEventById` switched from WebSocket to REST (`GET /api/event/{id}`)
- NIP-98 signing utility (`scripts/sign-nip98.mjs`)

## Key design decisions

- **Sync endpoint + cron dual-path.** Mobile gets near-instant confirmation via the sync endpoint. Non-Divine clients and missed events covered by the 1-minute cron. Both converge on `processKind5`.
- **NIP-98 author-only auth** on both endpoints. Caller pubkey must match kind 5 author.
- **IP rate limit runs before NIP-98 validation** to limit crypto work from unauthenticated callers.
- **D1 state taxonomy:** `accepted`, `success`, `failed:transient:*` (cron-retryable with exponential backoff), `failed:permanent:*` (terminal). Retries promote to `failed:permanent:max_retries_exceeded` at count 5.
- **Feature flag gates everything.** When off, routes 404 and cron is inert. Flippable without redeployment.
- **REST over WebSocket** for Funnelcake event lookups. Faster, works in Miniflare local dev, no CF Access header complexity.

## Test plan

705 tests pass (662 existing + 43 creator-delete + new validation tests from review fixes).

### Unit tests (done)
- [x] NIP-98 validation: 9 tests (happy path + 7 rejection paths including tags-array guard)
- [x] D1 helpers: 9 tests (claim idempotency + full decideAction state machine)
- [x] processKind5: 8 tests (happy path, multi-target, target_unresolved, no_sha256, transient 503, permanent 400, network error, idempotent skip)
- [x] Rate limiter: 2 tests (allow + block)
- [x] Sync endpoint: 8 tests (happy path + 400/401/403/404/429-IP/429-pubkey/202)
- [x] Status endpoint: 6 tests (happy path + 400/401/403/404/429)
- [x] Cron: 2 tests (happy path + transient retry with backoff)
- [x] Funnelcake fetch: 2 tests (retry-then-success + all-null)
- [x] Blossom client: 5 tests (action map + DELETE mapping + skip + error + network)
- [x] Existing index.test.mjs: 662 tests still pass (no regressions)

### Local e2e (done)
- [x] MinIO + Blossom (:7676) + moderation-service (:8787) all running locally
- [x] Feature flag gating: routes 404 when flag off, active when flag on
- [x] NIP-98 auth: rejects missing header (401), accepts valid signature
- [x] REST fetch from production Funnelcake: real kind 5 event fetched via `GET /api/event/{id}`
- [x] Pubkey authorization check: 403 on NIP-98/kind-5 pubkey mismatch (tested against real production kind 5)
- [x] Funnelcake read-after-write retry: 5 retries over 4.7s for non-existent event, returns 404
- [x] Structured observability log (`creator_delete.sync.request`) emitted
- [x] Moderation-service calls local Blossom and Blossom processes DELETE (verified via direct Blossom curl with webhook auth)

### Full pipeline e2e (deferred to production validation)
- [ ] processKind5 → Blossom call in a single request. Requires a real Divine test account with an uploaded video (ephemeral keys publish kind 34236 events that Funnelcake accepts at the protocol level but does not persist to ClickHouse). Each component is individually proven; the chain is validated during rollout plan steps 4-5 using a real test account.

### Code review (done)
- [x] Three rounds of self-review (comprehensive, fixes-focused, cross-repo integration contract)
- [x] All findings addressed: cron backoff, max_retries promotion, last-poll guard, IP rate-limit ordering, status-endpoint validation, stale row re-claim, path-traversal guard on eventId, unused imports removed, test mock URL corrected

### DIFF
--- a/docs/superpowers/plans/2026-04-16-per-video-delete-enforcement-plan.md
+++ b/docs/superpowers/plans/2026-04-16-per-video-delete-enforcement-plan.md
@@ -1,0 +1,2019 @@
+# Per-Video Delete End-to-End Enforcement Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Build the divine-moderation-service side of per-video creator-initiated deletes: a race-safe processing pipeline with a NIP-98 synchronous endpoint for divine-mobile's fast path and a 1-minute cron for non-Divine client coverage, writing to a D1 audit table and calling Blossom's new `DELETE` action.
+
+**Architecture:** Sync endpoint and cron both invoke a shared `processKind5` function that claims a D1 row via INSERT-OR-IGNORE, fetches the target event from Funnelcake (with read-after-write retry), extracts sha256, calls Blossom, records terminal state. State taxonomy distinguishes `failed:transient:*` (cron-retryable) from `failed:permanent:*`.
+
+**Tech Stack:** Cloudflare Workers, D1, KV, nostr-tools (for NIP-98 signature verification), vitest for tests.
+
+**Scope of this plan:** divine-moderation-service only (PR #92). Sibling work streams are:
+- **divine-blossom:** new `DELETE` action + cascade + `ENABLE_PHYSICAL_DELETE` flag + tombstone verification. Has its own plan when that work is picked up.
+- **divine-mobile:** polling + UI state machine. Its own plan, sibling of #3101.
+- **support-trust-safety:** one-line vocab doc update. Trivial; handled inline when Blossom PR lands.
+
+**Spec:** `docs/superpowers/specs/2026-04-16-per-video-delete-enforcement-design.md` on this branch.
+
+---
+
+## File Structure
+
+**New files:**
+
+- `migrations/006-creator-deletions.sql` — D1 schema for the audit table
+- `src/creator-delete/nip98.mjs` — NIP-98 Authorization header validation
+- `src/creator-delete/nip98.test.mjs`
+- `src/creator-delete/d1.mjs` — D1 helpers (claim, read-state, update-status)
+- `src/creator-delete/d1.test.mjs`
+- `src/creator-delete/process.mjs` — `processKind5` shared function
+- `src/creator-delete/process.test.mjs`
+- `src/creator-delete/rate-limit.mjs` — Per-pubkey + per-IP rate limiting via KV counters
+- `src/creator-delete/rate-limit.test.mjs`
+- `src/creator-delete/sync-endpoint.mjs` — `POST /api/delete/{kind5_id}` handler
+- `src/creator-delete/sync-endpoint.test.mjs`
+- `src/creator-delete/status-endpoint.mjs` — `GET /api/delete-status/{kind5_id}` handler
+- `src/creator-delete/status-endpoint.test.mjs`
+- `src/creator-delete/cron.mjs` — Scheduled work for kind 5 polling + transient retry
+- `src/creator-delete/cron.test.mjs`
+
+**Files to modify:**
+
+- `wrangler.toml` — Add a `* * * * *` cron entry alongside existing `*/5 * * * *`; verify existing `BLOSSOM_WEBHOOK_SECRET`, `BLOSSOM_ADMIN_URL`, `MODERATION_KV`, `BLOSSOM_DB` bindings are sufficient.
+- `src/index.mjs` — Register two routes on `moderation-api.divine.video`; dispatch new cron in the existing `scheduled(event, env, ctx)` handler based on `event.cron`.
+
+Each file has one clear responsibility. `process.mjs` is the hub; endpoints and cron orchestrate around it. Tests are colocated with source (existing repo pattern).
+
+---
+
+## Staging Preflight
+
+Complete these BEFORE starting implementation. Failures here can redirect the design.
+
+- [ ] **Funnelcake kind 5 queryability.** Open a WebSocket to `wss://relay.staging.divine.video`, send `["REQ","test",{"kinds":[5],"limit":5}]`, confirm events are returned (or that a recent kind 5 is visible). If kind 5s are treated as ephemeral and dropped, the cron strategy does not work and the design needs revision. Record result.
+
+    Command helper:
+    ```bash
+    wscat -c wss://relay.staging.divine.video
+    # after connect:
+    ["REQ","preflight",{"kinds":[5],"limit":5}]
+    ```
+
+- [ ] **Blossom webhook secret present in staging.** Confirm staging Blossom's `blossom_secrets` Fastly Secret Store has `webhook_secret` populated (same value moderation-service uses).
+
+    ```bash
+    # On staging Blossom:
+    fastly secret-store-entry list --store-id=<staging-store-id>
+    # Look for webhook_secret
+    ```
+
+- [ ] **Staging Blossom has PR #33 (or equivalent) deployed.** Confirms `Deleted` status is checked on HLS HEAD and subtitle-by-hash routes. Test: flip a staging blob to `Deleted` via `/admin/api/moderate` with action `BAN` (as a proxy since `DELETE` isn't wired yet), confirm the blob 404s on `/<sha256>`, `/<sha256>.jpg`, `/<sha256>.vtt`.
+
+- [ ] **D1 binding writable from staging.** Confirm the existing `BLOSSOM_DB` binding has write access from staging worker context.
+
+    ```bash
+    npx wrangler d1 execute blossom-webhook-events --env staging --command "SELECT name FROM sqlite_master WHERE type='table'"
+    ```
+
+- [ ] **NIP-98 verification path works end-to-end.** Sign a test NIP-98 header locally, send a request with it, confirm `nostr-tools` signature verification in the Worker accepts it. No existing code to reference; this is a first-principles check using `nostr-tools/pure` and `nostr-tools/nip98` helpers.
+
+If any preflight check fails, stop and address before writing implementation code.
+
+---
+
+## Task 1: D1 migration for creator_deletions
+
+**Files:**
+- Create: `migrations/006-creator-deletions.sql`
+
+- [ ] **Step 1: Write the migration SQL**
+
+    Create `migrations/006-creator-deletions.sql` with:
+
+    ```sql
+    -- Audit table for creator-initiated deletions (kind 5 events from Funnelcake).
+    -- Composite PRIMARY KEY ensures idempotency across concurrent invocations
+    -- (sync endpoint + cron colliding on the same kind 5).
+    --
+    -- status taxonomy:
+    --   accepted                              - claimed by a worker, in-progress
+    --   success                               - terminal success
+    --   failed:transient:{subcategory}        - retryable by cron (retry_count < 5)
+    --   failed:permanent:{subcategory}        - terminal, manual intervention required
+
+    CREATE TABLE IF NOT EXISTS creator_deletions (
+      kind5_id TEXT NOT NULL,
+      target_event_id TEXT NOT NULL,
+      creator_pubkey TEXT NOT NULL,
+      blob_sha256 TEXT,
+      status TEXT NOT NULL,
+      accepted_at TEXT NOT NULL,
+      completed_at TEXT,
+      retry_count INTEGER NOT NULL DEFAULT 0,
+      last_error TEXT,
+      PRIMARY KEY (kind5_id, target_event_id)
+    );
+
+    CREATE INDEX IF NOT EXISTS idx_creator_deletions_target ON creator_deletions(target_event_id);
+    CREATE INDEX IF NOT EXISTS idx_creator_deletions_creator ON creator_deletions(creator_pubkey);
+    CREATE INDEX IF NOT EXISTS idx_creator_deletions_sha256 ON creator_deletions(blob_sha256);
+    CREATE INDEX IF NOT EXISTS idx_creator_deletions_status ON creator_deletions(status);
+    ```
+
+- [ ] **Step 2: Apply migration to staging**
+
+    ```bash
+    npx wrangler d1 execute blossom-webhook-events --env staging --file migrations/006-creator-deletions.sql
+    ```
+
+    Expected: success, table created.
+
+- [ ] **Step 3: Verify table structure on staging**
+
+    ```bash
+    npx wrangler d1 execute blossom-webhook-events --env staging --command ".schema creator_deletions"
+    ```
+
+    Expected: output matches the CREATE TABLE statement.
+
+- [ ] **Step 4: Commit**
+
+    ```bash
+    git add migrations/006-creator-deletions.sql
+    git commit -m "feat: add creator_deletions audit table migration"
+    ```
+
+---
+
+## Task 2: NIP-98 validation module
+
+Validates an incoming NIP-98 Authorization header: base64-decoded kind 27235 event with `["u", url]` and `["method", method]` tags, `created_at` within ±60s, valid signature.
+
+**Files:**
+- Create: `src/creator-delete/nip98.mjs`
+- Create: `src/creator-delete/nip98.test.mjs`
+
+- [ ] **Step 1: Write the first failing test — valid NIP-98 header**
+
+    Create `src/creator-delete/nip98.test.mjs`:
+
+    ```javascript
+    import { describe, it, expect, beforeEach } from 'vitest';
+    import { generateSecretKey, getPublicKey, finalizeEvent } from 'nostr-tools/pure';
+    import { bytesToHex } from '@noble/hashes/utils';
+    import { validateNip98Header } from './nip98.mjs';
+
+    describe('validateNip98Header', () => {
+      let sk, pk;
+
+      beforeEach(() => {
+        sk = generateSecretKey();
+        pk = getPublicKey(sk);
+      });
+
+      function signNip98(url, method, skOverride) {
+        const event = finalizeEvent({
+          kind: 27235,
+          created_at: Math.floor(Date.now() / 1000),
+          tags: [['u', url], ['method', method]],
+          content: ''
+        }, skOverride || sk);
+        const encoded = btoa(JSON.stringify(event));
+        return `Nostr ${encoded}`;
+      }
+
+      it('accepts a valid signature for matching url and method', async () => {
+        const header = signNip98('https://moderation-api.divine.video/api/delete/abc123', 'POST');
+        const result = await validateNip98Header(header, 'https://moderation-api.divine.video/api/delete/abc123', 'POST');
+        expect(result.valid).toBe(true);
+        expect(result.pubkey).toBe(pk);
+      });
+    });
+    ```
+
+- [ ] **Step 2: Run test — confirm it fails**
+
+    ```bash
+    npx vitest run src/creator-delete/nip98.test.mjs
+    ```
+
+    Expected: FAIL with `Cannot find module './nip98.mjs'` or equivalent.
+
+- [ ] **Step 3: Implement the validator**
+
+    Create `src/creator-delete/nip98.mjs`:
+
+    ```javascript
+    // ABOUTME: NIP-98 HTTP Authorization header validation for creator-delete endpoints.
+    // ABOUTME: Validates base64-encoded kind 27235 event with u, method tags, ±60s clock drift, signature.
+
+    import { verifyEvent } from 'nostr-tools/pure';
+
+    const CLOCK_DRIFT_SECONDS = 60;
+    const EXPECTED_KIND = 27235;
+
+    export async function validateNip98Header(authorizationHeader, expectedUrl, expectedMethod) {
+      if (!authorizationHeader || !authorizationHeader.startsWith('Nostr ')) {
+        return { valid: false, error: 'Missing or malformed Authorization header (expected "Nostr <base64>")' };
+      }
+
+      const encoded = authorizationHeader.slice('Nostr '.length).trim();
+
+      let event;
+      try {
+        const decoded = atob(encoded);
+        event = JSON.parse(decoded);
+      } catch (e) {
+        return { valid: false, error: `Invalid base64 or JSON in Authorization header: ${e.message}` };
+      }
+
+      if (event.kind !== EXPECTED_KIND) {
+        return { valid: false, error: `Expected kind ${EXPECTED_KIND}, got ${event.kind}` };
+      }
+
+      const now = Math.floor(Date.now() / 1000);
+      if (Math.abs(now - event.created_at) > CLOCK_DRIFT_SECONDS) {
+        return { valid: false, error: `created_at ${event.created_at} outside ±${CLOCK_DRIFT_SECONDS}s window (server now: ${now})` };
+      }
+
+      const uTag = event.tags.find(t => t[0] === 'u')?.[1];
+      const methodTag = event.tags.find(t => t[0] === 'method')?.[1];
+
+      if (uTag !== expectedUrl) {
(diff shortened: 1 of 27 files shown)


rabble (commented):
Reviewed the full diff. Design is solid — nice separation, good race-safety around the D1 claim, sensible rate-limit ordering (IP before crypto). A few issues worth addressing before rollout, patch attached below.

## Pre-merge

### 🟡 `ctx.waitUntil` missing on budget-exceeded 202 path (`sync-endpoint.mjs:154-176`)
When the 8s budget trips, the handler returns 202 but `processing` isn't handed to `ctx.waitUntil`. In Workers runtime, unawaited promises are cancelled once the response is sent — so Blossom/D1 writes mid-flight get killed. Cron picks it up ~1 min later via the 30s stale-re-claim, but the creator eats that extra latency. `ctx` needs to be threaded from `fetch(req, env, ctx)` in `index.mjs` into the handler deps.

### 🟡 Empty-e-tags kind 5 returns spurious `success` 200 (`process.mjs:32-34`, `sync-endpoint.mjs:196-206`)
A signed kind 5 with no `e` tags produces `targets=[]`, no failure flag trips, endpoint returns `{status: 'success', targets: []}` with 200. Should 400 at the sync endpoint and not write any D1 rows.

### 🟡 No upper bound on e-tags per kind 5 (`process.mjs:430`)
NIP-09 allows N targets. `processKind5` iterates serially; each target is a Funnelcake fetch + Blossom call. A 1000-target kind 5 blows past the 8s budget and wastes the pubkey's rate-limit budget on failed ops. Cap at ~50 upfront.

### 🟡 Stale-row re-claim is not atomic (`process.mjs:455-459`)
After `claimRow` returns `{claimed: false, existing: stale}` and `decideAction` returns `'proceed'`, the unconditional `UPDATE SET accepted_at=NOW` lets two racing workers both re-claim the same row and duplicate Blossom work. Blossom DELETE is idempotent so final state is fine, but you'll see 2× POSTs and retry-count inflation. Fix: `UPDATE ... WHERE accepted_at = ?` with the stale timestamp, bail if `changes === 0`.

## Smaller items

- `rate-limit.mjs` comment says "sliding window" but it's fixed-window (bucketed by `floor(now/windowSeconds)`). Edge-of-bucket burst can hit 2× limit. Low abuse impact at 5/min pubkey, but the comment is misleading.
- `retry_count_after` log (`process.mjs:484, 506, 565`) reads from `claim.existing` which is pre-UPDATE — logs 0 when it should be `priorRetryCount + (isTransient ? 1 : 0)`. Just for observability correctness.
- `CREATOR_DELETE_RELAY_URL` isn't in the wrangler.toml secrets-docs block (referenced in `index.mjs:4143`).
- NIP-98 `u` tag uses strict string-equality against `url.toString()` — any scheme-case / trailing-slash drift between client and server breaks validation. Worth normalizing both sides.
- 8+ call sites of `console.log(JSON.stringify({...}))` for observability — easy `logEvent(obj)` helper extraction.
- Cron's `queryKind5Since` still uses WebSocket `queryRelay`, while PR description says "REST over WebSocket" — that only applies to `fetchNostrEventById` (point-lookup). Worth a comment explaining why range-query stays on WS.

## Security posture
NIP-98 + IP-before-crypto rate limit + feature flag default-off + path-traversal guard on event IDs: all good. `CF-Connecting-IP` is safe here (stripped at CF edge).

## Tests
43 new unit tests with strong coverage of state machine + rejection paths. Full-pipeline e2e deferred to rollout is reasonable given the ephemeral-key persistence gap noted in the PR.

Recommendation: approve after the four pre-merge items. Patch for those + the smaller fixes coming in a follow-up comment.

rabble:
Patch for the pre-merge items, 708/708 tests pass (added 3).

**Files touched:**
- `src/creator-delete/sync-endpoint.mjs` — add `ctx` dep, `ctx.waitUntil(processing)` on budget-exceeded path, reject kind 5 with no e-tags (400), extract `logRequest` helper
- `src/creator-delete/process.mjs` — `MAX_TARGETS_PER_KIND5 = 50` cap, atomic stale-row re-claim via conditional `UPDATE ... WHERE accepted_at = ?`, consolidated `priorRetryCount`
- `src/creator-delete/rate-limit.mjs` — comment correction (fixed-window, not sliding)
- `src/creator-delete/test-helpers.mjs` — fake D1 handles the 4-arg atomic re-claim UPDATE
- `src/index.mjs` — `fetch(request, env, ctx)` signature + pass `ctx` into handleSyncDelete deps
- Tests: 3 added (no e-tags → 400, ctx.waitUntil called, target cap)

**Not included** (flagged but minor / out of scope):
- NIP-98 `u` tag normalization
- `CREATOR_DELETE_RELAY_URL` secrets doc addition in wrangler.toml
- Cron `queryKind5Since` WS explanation comment

Patch (apply with `git apply pr-92-review-fixes.patch` from repo root):

```diff
src/creator-delete/process.mjs            | 47 +++++++++++----
 src/creator-delete/process.test.mjs       | 19 +++++-
 src/creator-delete/rate-limit.mjs         |  6 +-
 src/creator-delete/sync-endpoint.mjs      | 97 +++++++++++++------------------
 src/creator-delete/sync-endpoint.test.mjs | 28 +++++++++
 src/creator-delete/test-helpers.mjs       | 14 ++++-
 src/index.mjs                             |  3 +-
 7 files changed, 143 insertions(+), 71 deletions(-)

--- Changes ---

📄 src/creator-delete/process.mjs
  @@ -4,7 +4,12 @@
  -import { claimRow, updateToSuccess, updateToFailed, decideAction } from './d1.mjs';
  +import { claimRow, updateToSuccess, updateToFailed, decideAction, MAX_RETRY_COUNT } from './d1.mjs';
  +
  +// Cap per NIP-09 kind 5. One kind 5 with 1000 e-tags would blow past the
  +// sync-endpoint budget and burn Funnelcake/Blossom call budget. 50 is well
  +// above real-world deletes (single video, or a thread of ~handful).
  +export const MAX_TARGETS_PER_KIND5 = 50;
   
   /**
    * Extract the main blob sha256 from a kind 34236 video event.
  @@ -29,9 +34,20 @@
  -  const targetIds = (kind5.tags || [])
  +  const allTargetIds = (kind5.tags || [])
       .filter(t => t[0] === 'e' && t[1])
       .map(t => t[1]);
  +  const targetIds = allTargetIds.slice(0, MAX_TARGETS_PER_KIND5);
  +
  +  if (allTargetIds.length > MAX_TARGETS_PER_KIND5) {
  +    console.log(JSON.stringify({
  +      event: 'creator_delete.target_cap_applied',
  +      kind5_id: kind5.id,
  +      total_targets: allTargetIds.length,
  +      processed_targets: targetIds.length,
  +      trigger: triggerLabel
  +    }));
  +  }
   
     const resultTargets = [];
   
  @@ -59,11 +75,21 @@
  -    // action === 'proceed' — re-claim stale row by refreshing accepted_at
  +    // action === 'proceed'. If we didn't claim (row existed, stale/retryable),
  +    // conditionally re-claim: UPDATE only succeeds if accepted_at still matches
  +    // the value we observed, so two racing workers can't both reclaim.
       if (!claim.claimed) {
  -      await db.prepare(
  -        `UPDATE creator_deletions SET accepted_at = ?, status = 'accepted' WHERE kind5_id = ? AND target_event_id = ?`
  -      ).bind(acceptedIso, kind5.id, target_event_id).run();
  +      const reclaim = await db.prepare(
  +        `UPDATE creator_deletions
  +         SET accepted_at = ?, status = 'accepted'
  +         WHERE kind5_id = ? AND target_event_id = ? AND accepted_at = ?`
  +      ).bind(acceptedIso, kind5.id, target_event_id, claim.existing.accepted_at).run();
  +      const reclaimed = reclaim.meta.changes === 1 || reclaim.meta.rows_written === 1;
  +      if (!reclaimed) {
  +        // Another worker won the re-claim race. Report in_progress and let them finish.
  +        resultTargets.push({ target_event_id, status: 'in_progress' });
  +        continue;
  +      }
       }
       console.log(JSON.stringify({
         event: 'creator_delete.accepted',
  @@ -74,6 +100,8 @@
  +    const priorRetryCount = (claim.existing && claim.existing.retry_count) || 0;
  +
       const target = await fetchTargetEvent(target_event_id);
       if (!target) {
         await updateToFailed(db, {
  @@ -89,7 +117,7 @@
  -        retry_count_after: (claim.existing && claim.existing.retry_count) || 0,
  +        retry_count_after: priorRetryCount,
           trigger: triggerLabel
         }));
         resultTargets.push({ target_event_id, status: 'failed:permanent:target_unresolved' });
  @@ -111,7 +139,7 @@
  -        retry_count_after: (claim.existing && claim.existing.retry_count) || 0,
  +        retry_count_after: priorRetryCount,
           trigger: triggerLabel
         }));
         resultTargets.push({ target_event_id, status: 'failed:permanent:no_sha256' });
  @@ -148,11 +176,10 @@
  -    const priorRetryCount = (claim.existing && claim.existing.retry_count) || 0;
       const retryCountAfter = isTransient ? priorRetryCount + 1 : priorRetryCount;
   
       // Promote to permanent if retries exhausted
  -    const finalStatus = (isTransient && retryCountAfter >= 5)
  +    const finalStatus = (isTransient && retryCountAfter >= MAX_RETRY_COUNT)
         ? 'failed:permanent:max_retries_exceeded'
         : category;
   
  +37 -10

📄 src/creator-delete/process.test.mjs
  @@ -5,7 +5,7 @@
  -import { processKind5 } from './process.mjs';
  +import { processKind5, MAX_TARGETS_PER_KIND5 } from './process.mjs';
   import { makeFakeD1 } from './test-helpers.mjs';
   
   describe('processKind5', () => {
  @@ -118,4 +118,21 @@
  +
  +  it('caps processing at MAX_TARGETS_PER_KIND5 when given more e-tags', async () => {
  +    // Build a kind 5 with N+5 e-tags. Only the first MAX should be processed.
  +    const overflow = 5;
  +    const eTags = [];
  +    for (let i = 0; i < MAX_TARGETS_PER_KIND5 + overflow; i++) {
  +      eTags.push(['e', `t${i.toString().padStart(3, '0')}`]);
  +    }
  +    const kind5 = { id: 'k1', pubkey: 'pub1', tags: eTags };
  +    fetchTargetEvent.mockResolvedValue({ id: 't', pubkey: 'pub1', tags: [['imeta', `x ${SHA_C}`]] });
  +    callBlossomDelete.mockResolvedValue({ success: true, status: 200 });
  +
  +    const result = await processKind5(kind5, { db, fetchTargetEvent, callBlossomDelete });
  +
  +    expect(result.targets).toHaveLength(MAX_TARGETS_PER_KIND5);
  +    expect(callBlossomDelete).toHaveBeenCalledTimes(MAX_TARGETS_PER_KIND5);
  +  });
   });
  +18 -1

📄 src/creator-delete/rate-limit.mjs
  @@ -1,8 +1,10 @@
  -// ABOUTME: Simple KV-backed sliding window rate limiter for per-pubkey and per-IP limits.
  -// ABOUTME: Not perfectly accurate (no cross-region consistency) but sufficient for abuse prevention.
  +// ABOUTME: Simple KV-backed fixed-window rate limiter for per-pubkey and per-IP limits.
  +// ABOUTME: Bucketed by floor(now/windowSeconds) — edge-of-bucket bursts can briefly hit 2× limit.
  +// ABOUTME: No cross-region atomicity either. Sufficient for abuse prevention, not exact enforcement.
  +// ABOUTME: Use a Durable Object if strict limits become required.
   
   export async function checkRateLimit(kv, { key, limit, windowSeconds }) {
     const now = Math.floor(Date.now() / 1000);
  +4 -2

📄 src/creator-delete/sync-endpoint.mjs
  @@ -12,20 +12,25 @@
  +function logRequest(t0, kind5_id, status_code, extra = {}) {
  +  console.log(JSON.stringify({
  +    event: 'creator_delete.sync.request',
  +    status_code,
  +    latency_ms: Date.now() - t0,
  +    kind5_id: kind5_id || null,
  +    ...extra
  +  }));
  +}
  +
   export async function handleSyncDelete(request, deps) {
     const t0 = Date.now();
  -  const { db, kv, fetchKind5WithRetry, fetchTargetEvent, callBlossomDelete, budgetMs = 8000 } = deps;
  +  const { db, kv, ctx, fetchKind5WithRetry, fetchTargetEvent, callBlossomDelete, budgetMs = 8000 } = deps;
   
     const url = new URL(request.url);
     const kind5_id = url.pathname.split('/').pop();
   
     if (!kind5_id || !/^[a-f0-9]{64}$/i.test(kind5_id)) {
  -    console.log(JSON.stringify({
  -      event: 'creator_delete.sync.request',
  -      status_code: 400,
  -      latency_ms: Date.now() - t0,
  -      kind5_id: kind5_id || null
  -    }));
  +    logRequest(t0, kind5_id, 400);
       return jsonResponse(400, { error: 'Invalid kind5_id' });
     }
   
  @@ -33,12 +38,7 @@
  -    console.log(JSON.stringify({
  -      event: 'creator_delete.sync.request',
  -      status_code: 429,
  -      latency_ms: Date.now() - t0,
  -      kind5_id: kind5_id || null
  -    }));
  +    logRequest(t0, kind5_id, 429);
       return jsonResponse(429, {
         error: 'Rate limit exceeded',
         retry_after_seconds: ipCheck.retryAfterSeconds || 0
  @@ -47,24 +47,14 @@
  -    console.log(JSON.stringify({
  -      event: 'creator_delete.sync.request',
  -      status_code: 401,
  -      latency_ms: Date.now() - t0,
  -      kind5_id: kind5_id || null
  -    }));
  +    logRequest(t0, kind5_id, 401);
       return jsonResponse(401, { error: `NIP-98 validation failed: ${auth.error}` });
     }
   
     // Per-pubkey rate limit AFTER NIP-98 (pubkey only known after validation)
     const pubkeyCheck = await checkRateLimit(kv, { key: `pubkey:${auth.pubkey}`, limit: PER_PUBKEY_LIMIT, windowSeconds: RATE_WINDOW_SECONDS });
     if (!pubkeyCheck.allowed) {
  -    console.log(JSON.stringify({
  -      event: 'creator_delete.sync.request',
  -      status_code: 429,
  -      latency_ms: Date.now() - t0,
  -      kind5_id: kind5_id || null
  -    }));
  +    logRequest(t0, kind5_id, 429);
       return jsonResponse(429, {
         error: 'Rate limit exceeded',
         retry_after_seconds: pubkeyCheck.retryAfterSeconds || 0
  @@ -73,25 +63,22 @@
  -    console.log(JSON.stringify({
  -      event: 'creator_delete.sync.request',
  -      status_code: 404,
  -      latency_ms: Date.now() - t0,
  -      kind5_id: kind5_id || null
  -    }));
  +    logRequest(t0, kind5_id, 404);
       return jsonResponse(404, { error: 'Kind 5 not found on Funnelcake after retries' });
     }
   
     if (kind5.pubkey !== auth.pubkey) {
  -    console.log(JSON.stringify({
  -      event: 'creator_delete.sync.request',
  -      status_code: 403,
  -      latency_ms: Date.now() - t0,
  -      kind5_id: kind5_id || null
  -    }));
  +    logRequest(t0, kind5_id, 403);
       return jsonResponse(403, { error: 'Caller pubkey does not match kind 5 author' });
     }
   
  +  // Reject kind 5 with no e-tags upfront: produces no targets, would otherwise 200 with empty list
  +  const hasETag = (kind5.tags || []).some(t => t[0] === 'e' && t[1]);
  +  if (!hasETag) {
  +    logRequest(t0, kind5_id, 400, { reason: 'no_e_tags' });
  +    return jsonResponse(400, { error: 'Kind 5 event has no e-tags; nothing to delete' });
  +  }
  +
     const deadline = Date.now() + budgetMs;
     const processing = processKind5(kind5, {
  ... (truncated)
  @@ -100,16 +87,25 @@
  +  // When the budget elapses we return 202 and let processKind5 keep running.
  +  // ctx.waitUntil keeps the Worker alive past the Response; without it the
  +  // runtime cancels the promise mid-Blossom-call and recovery waits on the
  +  // next cron tick (~1 min latency).
  +  if (ctx && typeof ctx.waitUntil === 'function') {
  +    ctx.waitUntil(processing.catch(err => {
  +      console.error(JSON.stringify({
  +        event: 'creator_delete.sync.waituntil_error',
  +        kind5_id,
  +        error: err?.message
  +      }));
  +    }));
  +  }
  +
     const timeoutPromise = new Promise(resolve => setTimeout(() => resolve({ budgetExceeded: true }), budgetMs));
     const raceResult = await Promise.race([processing, timeoutPromise]);
   
     if (raceResult.budgetExceeded) {
  -    console.log(JSON.stringify({
  -      event: 'creator_delete.sync.request',
  -      status_code: 202,
  -      latency_ms: Date.now() - t0,
  -      kind5_id: kind5_id || null
  -    }));
  +    logRequest(t0, kind5_id, 202);
       return jsonResponse(202, {
         kind5_id,
         status: 'in_progress',
  @@ -121,13 +117,7 @@
  -    // One target still had an in-progress existing row. Return 202.
  -    console.log(JSON.stringify({
  -      event: 'creator_delete.sync.request',
  -      status_code: 202,
  -      latency_ms: Date.now() - t0,
  -      kind5_id: kind5_id || null
  -    }));
  +    logRequest(t0, kind5_id, 202);
       return jsonResponse(202, {
         kind5_id,
         status: 'in_progress',
  @@ -135,12 +125,7 @@
  -  console.log(JSON.stringify({
  -    event: 'creator_delete.sync.request',
  -    status_code: 200,
  -    latency_ms: Date.now() - t0,
  -    kind5_id: kind5_id || null
  -  }));
  +  logRequest(t0, kind5_id, 200);
     return jsonResponse(200, {
       kind5_id,
       status: anyFailed ? 'failed' : 'success',
  +41 -56

📄 src/creator-delete/sync-endpoint.test.mjs
  @@ -135,4 +135,32 @@
  +
  +  it('returns 400 when kind 5 has no e-tags', async () => {
  +    const kind5 = { id: KIND5_ID, pubkey: pk, tags: [['p', pk]] }; // no e-tag
  +    deps.fetchKind5WithRetry.mockResolvedValueOnce(kind5);
  +    const url = `https://moderation-api.divine.video/api/delete/${KIND5_ID}`;
  +    const request = new Request(url, { method: 'POST', headers: { Authorization: signNip98(url, 'POST') } });
  +    const response = await handleSyncDelete(request, deps);
  +    expect(response.status).toBe(400);
  +    const body = await response.json();
  +    expect(body.error).toMatch(/no e-tags/i);
  +    expect(deps.fetchTargetEvent).not.toHaveBeenCalled();
  +    expect(deps.callBlossomDelete).not.toHaveBeenCalled();
  +  });
  +
  +  it('calls ctx.waitUntil with the processing promise when budget is exceeded', async () => {
  +    const kind5 = { id: KIND5_ID, pubkey: pk, tags: [['e', 't1']] };
  +    deps.fetchKind5WithRetry.mockResolvedValueOnce(kind5);
  +    deps.fetchTargetEvent.mockResolvedValueOnce({ id: 't1', pubkey: pk, tags: [['imeta', `x ${SHA_C}`]] });
  +    deps.callBlossomDelete.mockReturnValueOnce(new Promise(() => {})); // never resolves
  +
  +    const ctx = { waitUntil: vi.fn() };
  +    const url = `https://moderation-api.divine.video/api/delete/${KIND5_ID}`;
  +    const request = new Request(url, { method: 'POST', headers: { Authorization: signNip98(url, 'POST') } });
  +    const response = await handleSyncDelete(request, { ...deps, ctx, budgetMs: 50 });
  +    expect(response.status).toBe(202);
  +    expect(ctx.waitUntil).toHaveBeenCalledTimes(1);
  +    expect(ctx.waitUntil.mock.calls[0][0]).toBeInstanceOf(Promise);
  +  });
   });
  +28 -0

📄 src/creator-delete/test-helpers.mjs
  @@ -25,10 +25,22 @@
  +            // Atomic re-claim: UPDATE ... WHERE kind5_id=? AND target_event_id=? AND accepted_at=?
  +            // binds: [new_accepted_at, kind5_id, target_event_id, old_accepted_at]
  +            if (this._sql.includes('AND accepted_at = ?')) {
  +              const [newAccepted, kind5_id, target_event_id, oldAccepted] = this._binds;
  +              const target_key = `${kind5_id}:${target_event_id}`;
  +              const existing = rows.get(target_key);
  +              if (!existing || existing.accepted_at !== oldAccepted) {
  +                return { meta: { changes: 0, rows_written: 0 } };
  +              }
  +              rows.set(target_key, { ...existing, accepted_at: newAccepted, status: 'accepted' });
  +              return { meta: { changes: 1, rows_written: 1 } };
  +            }
  +            // Other UPDATEs: kind5_id and target_event_id are the last two binds.
               const target_key = `${this._binds[this._binds.length - 2]}:${this._binds[this._binds.length - 1]}`;
               const existing = rows.get(target_key);
               if (existing) {
  -              // Very simplified: we're just testing the wrappers, not SQL correctness.
                 rows.set(target_key, { ...existing, _updated: true });
                 return { meta: { changes: 1 } };
               }
  +13 -1

📄 src/index.mjs
  @@ -2026,7 +2026,7 @@
  -  async fetch(request, env) {
  +  async fetch(request, env, ctx) {
       const url = new URL(request.url);
       const startTime = Date.now();
       const requestId = crypto.randomUUID().substring(0, 8);
  @@ -4154,6 +4154,7 @@
  +          ctx,
             fetchKind5WithRetry: (id) => fetchKind5WithRetry(id, {
               fetchEventById: (eid) => fetchNostrEventById(eid, [relayUrl], env)
             }),
  +2 -1
```

NotThatKindOfDrLiz (approved):
Approved. I rechecked the live PR state after the follow-up fixes: no unresolved review threads, CI is green, the branch is mergeable, and the full local suite passed (779 tests).

The original pre-merge concerns are addressed: ctx.waitUntil keeps budget-exceeded sync work alive, no-e-tag kind 5 events return 400, kind 5 target processing is capped, and stale-row reclaim is atomic. The later cleanup items are also handled or tracked: retry constants are centralized, the in-progress timeout/idempotency nuance is documented, CREATOR_DELETE_RELAY_URL is documented, status queries now have pre-auth IP limiting, and the missing guard tests are in place.

Remaining work is correctly follow-up scoped and should not block this PR: NIP-98 URL normalization (#100), real full-pipeline e2e validation (#101), and Blossom DELETE idempotency docs (#102/#103). Keep CREATOR_DELETE_PIPELINE_ENABLED=false until the rollout validation steps are ready.

PR #97, opened 2026-04-17
feat: visible-watermark detector for AI-generated video

rabble:
## Summary
- Adds `detectLogos(frames)` which crops four 15% corners per frame (TL/TR/BL/BR) and runs a per-crop classifier for Meta sparkle, Sora, Veo, Runway, Kling, Pika, Luma, + `clean`/`other_logo`.
- Adds `aggregateLogoDetections(detections)` — majority vote per `(corner, class)` at confidence ≥ 0.7, triggers at ≥ 50% frames. One-off false positives are discarded.
- Model inference is stubbed behind `loadModel()` / `runInference()` so the ONNX model (via `onnxruntime-web`) can swap in once trained. `LOGO_DETECTOR_MODEL_URL` wired in `wrangler.toml`.
- Documents the class→generator mapping in `CONTENT_MODERATION.md` and notes that `meta_sparkle` is our primary Meta-AI signal only until Meta's Video Seal invisible watermark is decodable.

## Design notes
- `runInference()` currently returns `{class: 'clean', confidence: 1.0}` so the pipeline runs deterministically in dev and tests until the ONNX file ships.
- `detectLogos(frames, { infer, model })` accepts an injected `infer` for tests and for the future ONNX swap; no behavioural seam change needed when the real model lands.
- Aggregator tie-break: first by frame count, then by average confidence within tied classes, so a stronger but smaller-corner vote can't displace a broader one.

## Test plan
- [x] `npx vitest run src/moderation/logo_detector.test.mjs src/moderation/logo_aggregator.test.mjs` — **27/27 passing**
- [x] TDD: tests written and verified failing before either module existed.
- [x] Full suite: no new regressions (4 pre-existing failures on `main` are unrelated transcript + original-vine tests).
- [ ] Follow-up PR will wire `detectLogos` into `src/moderation/pipeline.mjs` once the ONNX model is trained and hosted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CONTENT_MODERATION.md
+++ b/CONTENT_MODERATION.md
@@ -536,6 +536,51 @@ Track these metrics:
 - API costs
 - Manual review queue size
 
+## Visible-Watermark Detector (AI-generated video)
+
+Most consumer AI video generators stamp a small, visually distinctive mark in a
+fixed corner of every frame. A narrow CNN classifier on 15% corner crops is
+highly effective and runs inside the Worker via `onnxruntime-web`.
+
+Source files:
+
+- `src/moderation/logo_detector.mjs` — extracts TL/TR/BL/BR crops per frame and
+  runs the classifier. `loadModel()` / `runInference()` are stubs today; they
+  become real ONNX calls once `LOGO_DETECTOR_MODEL_URL` is set and the model is
+  trained.
+- `src/moderation/logo_aggregator.mjs` — majority vote across frames. A verdict
+  is emitted when ≥50% of frames flag the **same class in the same corner** at
+  confidence ≥0.7. One-off false positives are discarded by design.
+
+### Class → generator mapping
+
+| Class           | Generator                            | Typical mark                            |
+|-----------------|--------------------------------------|-----------------------------------------|
+| `clean`         | —                                    | No visible AI watermark                 |
+| `meta_sparkle`  | Meta Imagine / Movie Gen             | Four-point sparkle icon, lower-left     |
+| `openai_sora`   | OpenAI Sora                          | Moving Sora wordmark                    |
+| `google_veo`    | Google Veo                           | "Veo" text watermark                    |
+| `runway`        | Runway (Gen-2/Gen-3)                 | Corner Runway logo                      |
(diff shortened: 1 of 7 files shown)


rabble:
Review response:

**Moving-watermark miss (Sora)** — correct catch. My own docs said "Moving Sora wordmark" but the aggregator required same-corner agreement, so it silently rejected every moving-watermark frame. Fixed in 3fb5776: added a class-only fallback pass. Static `(corner, class)` match still wins over a scattered class-only match when both qualify — a same-corner logo is a tighter signal than a hop-scattered one. Test `detects moving watermarks that hop corners across frames (Sora case)` pins the new behavior, and `prefers a static over moving when both qualify` pins the precedence.

**Mixed-confidence coverage gap** — added test `averages confidence only over flags that cleared the 0.7 floor`. Below-floor detections are excluded from both the frame count and the confidence average.

**`loadModelFromEnv` over-tested** — keeping as-is. Four small tests on a 4-line helper is mild, and collapsing to a table-driven test is churn without meaningful improvement.

**URL validation for ONNX fetch** — noted for the wiring PR. When `loadModel()` actually fetches, scheme + host-allowlist validation goes in before the fetch.

**`detectLogos` injection seam** — confirmed forward-compatible with ONNX. No change needed.

34 tests passing (was 31). CI green.

NotThatKindOfDrLiz (approved):
Approved. I did a final pass on the live head and found no blocking issues.

What I checked:
- PR is mergeable; only blocked by required review.
- CI is green: Lint/Test/CLA pass, Deploy skipped.
- No open review threads.
- The new detector is not wired into the runtime moderation pipeline yet; references are limited to docs/config plus the new detector and aggregator modules/tests.
- Focused logo detector/aggregator tests pass: 34 tests.
- Full suite passes locally: 716 tests.
- git diff --check is clean.

One non-blocking calibration note for the future wiring/model PR: the aggregator intentionally lets a qualifying static corner/class winner take precedence over the moving class-only winner. That is documented and currently tested as the desired behavior, and since this PR is disabled-by-default scaffolding rather than live policy, I do not think it should block merge. Revisit that winner policy once real model outputs are available.

PR #98, opened 2026-04-17
feat: interpret and persist Video Seal signals

rabble:
## Summary
- accept upstream Video Seal queue fields and interpret payloads in the moderation pipeline
- persist the interpreted `videoseal` signal in D1 and expose it from admin/public moderation lookups
- add schema validation, tests, docs, and the D1 migration for the new signal

## Test Plan
- [x] npm test
- [x] npx wrangler d1 execute blossom-webhook-events --remote --command "PRAGMA table_info(moderation_results);"

### DIFF
--- a/CONTENT_MODERATION.md
+++ b/CONTENT_MODERATION.md
@@ -28,6 +28,14 @@ graph TD
 2. **Multi-stage**: Quick checks first, expensive analysis later
 3. **Fail-safe**: Content remains accessible unless definitively harmful
 
+## Video Seal
+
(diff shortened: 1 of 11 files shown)


NotThatKindOfDrLiz (approved):
Approved. I did one final pass after rebasing this onto current main and resolving the C2PA/Video Seal overlap. The migration is now correctly numbered as 007 after the creator-delete migration, admin lookup preserves both C2PA raw_response data and persisted videoseal metadata, and the signed-AI short-circuit still returns the stable videoseal shape.\n\nI also added a small comment documenting why unknown Video Seal prefixes are downweighted until calibrated. Local full suite passes (58 files / 921 tests), and GitHub CI is green for Lint, Test, and CLA.

PR #104, opened 2026-04-17
fix(nip98): normalize URLs before comparing u tag to expected URL

### LINKED ISSUE #100: Normalize NIP-98 URL comparison for creator-delete endpoints
## Context
PR #92 added NIP-98-gated creator-delete endpoints. The current validation compares the signed `u` tag and request URL as raw strings, which means semantically equivalent URL forms can fail auth.

## Work
- Define canonicalization rules for NIP-98 URL comparison in this service.
- Normalize both the signed `u` tag and the request URL before comparing, likely via `new URL(...).toString()` unless we decide stricter rules are required.
- Add tests covering trailing slash behavior, explicit default ports, and percent-encoding/canonical path behavior.
- Verify existing NIP-98 replay-window and method checks remain unchanged.

## Notes
This is auth-adjacent and should be handled as a small isolated PR.

mbradley:
Closes #100.

Stacked on #92 (`spec/per-video-delete-enforcement`) because `nip98.mjs` lives there. Do not merge until #92 lands.

## Problem

The existing NIP-98 validator compared the signed `u` tag against the server-computed expected URL with raw string equality. Semantically equivalent forms produced false negatives:

- `https://host:443/path` vs `https://host/path`
- `HTTPS://Host/path` vs `https://host/path`
- `https://host` vs `https://host/`
- `https://host/path#frag` vs `https://host/path`

Any well-behaved client that emitted one form while the server computed the other would fail auth.

## Fix

Added `normalizeUrl` helper that runs both sides through `new URL(...).toString()` and explicitly strips fragment and userinfo. Comparison runs on the normalized forms.

### What normalizes
- **Via the WHATWG URL constructor** (stable across all JS runtimes that host CF Workers): lowercase scheme and host; strip default ports (`:443` on https, `:80` on http); normalize bare root (`https://x`) to `https://x/`.
- **Explicit in the helper**: strip fragment (never reaches server); strip userinfo (API auth URLs never carry credentials; see inline comment for the security rationale).

### What deliberately does NOT normalize (documented in the helper comment)
- Non-root trailing slash (`/foo` vs `/foo/` are distinct resources by HTTP convention).
- Query parameter ordering.
- Percent-encoding of reserved/unreserved characters.

Signer and verifier must agree on these dimensions. The URL constructor is the natural tool on both sides.

## Unchanged

Replay-window (`±60s` clock drift) and method comparison run exactly as before. Signature verification still runs on the original event — canonicalization only affects the equality check, not the cryptographic verification.

## Tests

19 new test cases across two describe blocks:

- `URL canonicalization` (10 tests) exercises the full validator: default ports, uppercase scheme/host, root slash, fragment, non-root trailing slash staying distinct, different paths, different hosts, unparseable u tag. Includes explicit guards that replay-window and method-mismatch rejections still fire unchanged.
- `normalizeUrl helper` (9 tests) unit-tests the helper directly across the canonicalization rules and the null-on-unparseable contract.

All 70 tests in the creator-delete module pass locally (`npx vitest run src/creator-delete/`).

## Review pass

Ran a code-review pass on the diff before opening. Key findings addressed in a follow-up commit:
- Docstring attribution corrected: the URL constructor (not the helper) does default-port stripping, scheme/host lowercasing, and bare-root-slash normalization.
- Added inline comment at the userinfo-stripping lines noting the security tradeoff — signer is bound to the original event, not the normalized form used only for equality.

## Verification

Every claim was cross-checked against `spec/per-video-delete-enforcement` at `bcf2a1f`.

### DIFF
--- a/src/creator-delete/nip98.mjs
+++ b/src/creator-delete/nip98.mjs
@@ -9,6 +9,44 @@ import { verifyEvent } from 'nostr-tools/pure';
 const CLOCK_DRIFT_SECONDS = 60;
 const EXPECTED_KIND = 27235;
 
+/**
+ * Canonicalize a URL for NIP-98 `u` tag comparison. Normalizes forms that are
+ * semantically equivalent but textually distinct so that a signed `u` tag from
+ * a well-behaved client matches the request URL we build server-side.
+ *
+ * What normalizes:
+ * - Scheme and host case (`HTTPS://Example.COM` → `https://example.com`).
+ * - Explicit default ports (`https://x:443/path` → `https://x/path`,
+ *   `http://x:80/path` → `http://x/path`).
+ * - Root path (`https://x` → `https://x/`).
+ * - Fragment stripped (fragments never reach the server).
+ * - Userinfo stripped (not expected on API auth URLs; normalized away so a
+ *   stray `user@host` form doesn't mismatch).
+ *
+ * What does NOT normalize (documented behavior — signer and verifier must agree):
+ * - Non-root trailing slash (`/path` vs `/path/` are distinct resources).
+ * - Query parameter ordering.
+ * - Percent-encoding of reserved or unreserved characters (the URL constructor
+ *   preserves the encoding as-received). If a signer produces `%7E` and the
+ *   request URL has `~`, they will not match. Clients should build both sides
+ *   with the same serialization; the typical path is `new URL(x).toString()`.
+ *
+ * Returns `null` if the input cannot be parsed as a URL. Callers must treat
+ * `null === null` as a mismatch, not a match.
+ */
+function normalizeUrl(url) {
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz (approved):
Approved. I did a final review after the branch cleanup: the live diff is scoped to src/creator-delete/nip98.mjs and src/creator-delete/nip98.test.mjs, there are no comments or unresolved review threads, CI is green, and the focused creator-delete suite passed locally (70 tests).

The normalization is appropriately conservative for NIP-98 URL comparison: it accepts semantically equivalent forms such as default ports, scheme/host case, root slash, and fragments, while leaving path slash semantics, query ordering, and percent-encoding differences strict. The userinfo stripping is also acceptable here because the signature remains bound to the original event/pubkey; normalization is only used for equality against the expected request URL.

PR #106, opened 2026-04-18
feat(sweep): creator-delete validation-window physical-delete sweep

### LINKED ISSUE #90: Create validation-window physical-delete sweep for creator-deleted blobs
Follow-up from PR #85 review. This should not block PR #85 once the requested correctness fixes land.

## Context

The rollout plan intentionally starts with `ENABLE_PHYSICAL_DELETE=false`. During that validation window, creator-delete requests will soft-delete blobs but leave bytes in GCS.

## Scope

Create a one-time/admin sweep that can physically delete validation-window blobs after the flag is enabled.

Suggested behavior:

- Identify creator-deleted blobs from the validation window.
- Verify Blossom metadata status is `Deleted` before destroying bytes.
- Reuse the same physical-delete helper/policy path where practical.
- Log per-blob success/failure with enough detail for operator follow-up.
- Support dry-run mode before destructive execution.

## Acceptance criteria

- Dry run lists target blobs without deleting.
- Execution physically removes main GCS bytes and eligible artifacts.
- Failures do not stop the entire sweep, but are summarized at the end.

mbradley:
Closes divinevideo/divine-blossom#90.

Spec: `docs/superpowers/specs/2026-04-17-creator-delete-validation-sweep-design.md`
Plan: `docs/superpowers/plans/2026-04-17-creator-delete-validation-sweep-plan.md`

## Summary

Operator-run script that backfills physical byte deletion for creator-delete kind 5 events whose Blossom soft-delete completed during the `ENABLE_PHYSICAL_DELETE=false` validation window.

- Migration 007 adds `physical_deleted_at TEXT` to `creator_deletions`.
- New script `scripts/sweep-creator-deletes.mjs`: reads candidates from D1, calls Blossom DELETE via the existing `notifyBlossom()` client (request shape identical to live pipeline), stamps the column on confirmed byte destruction.
- Pre-flight aborts (exit 2) if `ENABLE_PHYSICAL_DELETE` is off, auth fails, or Blossom is unreachable. Zero D1 writes in those cases.
- Strict success criterion: only stamps when Blossom returns `physical_deleted: true`. A 200 with `physical_deleted: false` is treated as failure (row stays unstamped, retried on next run).
- Surfaces unprocessable rows (status=success, NULL sha) and permanent failures (status=failed:permanent:*) in the summary so creator intent we cannot fulfill is loud, not silent.
- Per-row JSONL outcome lines on stdout for grep/jq.
- Best-effort SIGINT drain (first Ctrl-C drains in-flight + flushes; second Ctrl-C exits 130).
- `--local` flag swaps `--remote` for `--local` on wrangler calls so the same script runs against local D1 for smoke testing.

## Test plan

- [x] Unit + integration tests via vitest (72 new test cases; full suite 905 / 905)
- [x] Lint clean
- [x] Local smoke: migration applied to local D1, dry-run found a synthetic success row, exited 0
- [ ] Apply migration to prod D1 (operator step at rollout)
- [ ] Dry-run against prod D1 (operator step at rollout)
- [ ] First real sweep run (deferred to rollout — see `support-trust-safety/docs/rollout/2026-04-16-creator-delete-rollout.md`)

## Notes for reviewers

- The script reuses `src/blossom-client.mjs::notifyBlossom` to guarantee identical Blossom request shape to the live creator-delete pipeline. No new HTTP surface.
- `node:child_process` is imported dynamically inside `defaultRunner` so the Workers test pool (which does not provide that module even with `nodejs_compat`) does not fail on module collection.
- Output uses `console.log` rather than `process.stdout.write` for the same Workers-compat reason. Per-line behavior is unchanged.
- SQL is string-interpolated because `wrangler d1 execute --command` does not accept bind parameters. Inputs are validated before reaching the SQL builder: sha256 must be `/^[0-9a-f]{64}$/`, timestamps must round-trip through `new Date().toISOString()`, limits must be non-negative integers.

### DIFF
--- a/scripts/sweep-creator-deletes.mjs
+++ b/scripts/sweep-creator-deletes.mjs
@@ -151,3 +151,44 @@ export async function runWithConcurrency(items, concurrency, fn) {
   await Promise.all(workers);
   return results;
 }
+
+import { notifyBlossom as defaultNotify } from '../src/blossom-client.mjs';
+
+/**
+ * Wraps notifyBlossom() so the script reuses the live-pipeline request shape and headers.
+ * notifyImpl is injectable for tests.
+ */
+export async function callBlossomDelete(sha256, cfg, notifyImpl = defaultNotify) {
+  const env = {
+    BLOSSOM_WEBHOOK_URL: cfg.blossomWebhookUrl,
+    BLOSSOM_WEBHOOK_SECRET: cfg.blossomWebhookSecret
+  };
+  const r = await notifyImpl(sha256, 'DELETE', env);
+  if (r.success) {
+    return { ok: true, status: r.status, body: r.result };
+  }
+  return {
+    ok: false,
+    status: r.status,
+    networkError: !!r.networkError,
+    error: r.error
+  };
+}
+
+/**
+ * Classifies a Blossom call result into the action the script should take.
+ * Used by both pre-flight and per-row sweep logic.
+ */
+export function classifyDeleteResult(r) {
+  if (r.ok) {
+    const b = r.body || {};
+    if (b.physical_delete_enabled === false) return { kind: 'flag-off' };
+    if (b.status === 'error') return { kind: 'failure', reason: b.error || 'blossom returned status=error' };
+    if (b.status === 'success' && b.physical_deleted === true) return { kind: 'success' };
+    return { kind: 'failure', reason: 'physical_deleted=false despite flag on' };
+  }
+  if (r.status === 401 || r.status === 403) return { kind: 'auth-failure' };
+  if (r.networkError) return { kind: 'unreachable', reason: r.error || 'network error' };
+  if (r.status >= 500) return { kind: 'unreachable', reason: `HTTP ${r.status}` };
+  return { kind: 'failure', reason: r.error || `HTTP ${r.status}` };
+}
--- a/scripts/sweep-creator-deletes.mjs
+++ b/scripts/sweep-creator-deletes.mjs
@@ -1,0 +1,65 @@
+#!/usr/bin/env node
+// This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0.
+// If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// ABOUTME: Validation-window physical-delete sweep for creator-deleted blobs (blossom#90).
+// ABOUTME: Reads creator_deletions from D1, asks Blossom to destroy bytes, stamps physical_deleted_at.
+
+const DEFAULT_BLOSSOM_WEBHOOK_URL = 'https://media.divine.video/admin/moderate';
+const DEFAULT_D1_DATABASE = 'divine-moderation-decisions-prod';
+const DEFAULT_CONCURRENCY = 5;
+const FLUSH_BATCH_SIZE = 100;
+const SHA256_HEX = /^[0-9a-f]{64}$/;
+
+function getFlag(argv, name) {
+  const prefix = `--${name}=`;
+  for (const a of argv) {
+    if (a === `--${name}`) return true;
+    if (a.startsWith(prefix)) return a.slice(prefix.length);
+  }
+  return null;
+}
+
+function validateIso(value, fieldName) {
+  if (value == null) return null;
+  const d = new Date(value);
+  if (Number.isNaN(d.getTime())) {
+    throw new Error(`Invalid ${fieldName}: ${value} (must be ISO 8601)`);
+  }
+  return d.toISOString();
+}
+
+function validatePositiveInt(value, fieldName) {
+  const n = Number(value);
+  if (!Number.isInteger(n) || n <= 0) {
+    throw new Error(`Invalid ${fieldName}: ${value} (must be positive integer)`);
+  }
+  return n;
+}
+
+function validateNonNegativeInt(value, fieldName) {
+  const n = Number(value);
+  if (!Number.isInteger(n) || n < 0) {
+    throw new Error(`Invalid ${fieldName}: ${value} (must be non-negative integer)`);
+  }
+  return n;
+}
+
+export function parseArgs(argv) {
+  const dryRun = getFlag(argv, 'dry-run') === true;
+  const since = validateIso(getFlag(argv, 'since') || null, 'since');
+  const until = validateIso(getFlag(argv, 'until') || null, 'until');
+
+  const rawConcurrency = getFlag(argv, 'concurrency');
+  const concurrency = rawConcurrency
+    ? validatePositiveInt(rawConcurrency, 'concurrency')
+    : DEFAULT_CONCURRENCY;
+
+  const rawLimit = getFlag(argv, 'limit');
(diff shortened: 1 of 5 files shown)


mbradley:
## Self-review pass (code-reviewer agent)

Ran the feature-dev code-reviewer agent on the diff. Findings addressed in 0d1f817:

**Accepted and fixed:**
- **#2 (Important, 95% confidence):** `flushDeletedAt` errors mid-sweep fell through to the outer catch as exit 99 with a bare stack trace. Spec said exit 4 with the sha list. Added `D1WriteAbort` error class carrying `unflushedShas`; `main()` catches it, prints the shas with retry guidance, returns 4. Same treatment for the pre-flight stamp call. New unit + integration tests cover both paths.
- **#3 (Minor, 82% confidence):** Pre-flight row stamped after bulk-sweep flush, not "alongside" as the spec said. Added code comment documenting the retry-safety rationale (Blossom DELETE on already-gone bytes returns `physical_deleted: true` per Blossom PR #85, so a second run stamps it).
- **#4 (Theoretical, 80% confidence):** `classifyDeleteResult` ordering keeps `flag-off` before `status=error` because flag-off is the more actionable operator signal in the one realistic case. Added comment explaining the reasoning.

**Rejected (false positive):**
- **#1 (Critical, 100% confidence):** Agent flagged `DEFAULT_D1_DATABASE = 'blossom-webhook-events'` as the deprecated binding. Verified against `wrangler.toml:26-29` — `BLOSSOM_DB → database_name = blossom-webhook-events` is the active mod-service D1 (where `creator_deletions` actually lives, per every query in `src/creator-delete/*.mjs`). The agent confused this with the deprecated `blossom-webhook-events` Cloudflare Worker (different thing, same name — listed under the Relay Manager Legacy row in CLAUDE.md).

Full test suite green (907/907), lint clean.

scripts/sweep-creator-deletes.mjs
```diff
@@ -0,0 +1,615 @@
...
+      throw new D1WriteAbort([...pending], e);
+    }
+  }
+
+  const results = await runWithConcurrency(candidates, cfg.concurrency, async (row) => {
+    return callBlossomDelete(row.blob_sha256, cfg, notifyImpl);
+  }, isDraining);
+
+  for (const r of results) {
+    if (isDraining()) break;  // SIGINT: stop scheduling new work, flush pending, exit
```
NotThatKindOfDrLiz: This `break` undercuts the promised drain behavior. `runWithConcurrency(..., isDraining)` has already awaited the in-flight work and returned the subset of results that actually ran. Breaking here means any successes that completed after the first SIGINT are dropped before they can be added to `pending` and flushed, so the partial summary can miss bytes that were actually deleted. I think the drain check belongs in the scheduler only, not in this result-consumption loop.

NotThatKindOfDrLiz: This 0break0 undercuts the promised drain behavior. 0runWithConcurrency(..., isDraining)0 has already awaited the in-flight work and returned the subset of results that actually ran. Breaking here means any successes that completed after the first SIGINT are dropped before they can be added to 0pending0 and flushed, so the partial summary can miss bytes that were actually deleted. I think the drain check belongs in the scheduler only, not in this result-consumption loop.

scripts/sweep-creator-deletes.mjs
```diff
@@ -0,0 +1,615 @@
...
+  }
+
+  // Surfacing queries
+  let unprocessable = [];
+  let permanentFailures = [];
+  try {
+    unprocessable = await fetchUnprocessable(cfg, runner);
+    permanentFailures = await fetchPermanentFailures(cfg, runner);
+  } catch (e) {
+    console.error(`surfacing query failed: ${e.message}`);
```
NotThatKindOfDrLiz, line 586: I do not think this should degrade to a logged warning. If these surfacing queries fail, the operator loses the visibility the script is supposed to provide for unprocessable and permanent-failure rows, but the current code can still return exit 0 because both lists stay empty. That makes the run look clean when the final readback step actually failed. I would treat this as a non-zero outcome and add a test that locks that in.

NotThatKindOfDrLiz (changes requested):
The overall shape here is solid, and I agree with the strict success criterion and the reuse of the existing Blossom client. I am still at request changes for two operator-path correctness issues before I would rely on this in production.\n\n1. The SIGINT drain path does not currently preserve completed in-flight results. The concurrency helper already returns the subset of items that actually ran, including work that finished after the first signal, but the post-processing loop breaks immediately when draining is set. That means completed successes after the signal are never added to pending, never flushed, and never reflected in the partial summary.\n\n2. A failure in the post-sweep surfacing queries can still produce exit code 0. Right now the code logs a surfacing-query failure and then computes the final exit code from empty unprocessable and permanentFailures lists. For an operator-run sweep whose explicit goal is to make unfulfilled creator intent loud, that should be treated as a non-zero outcome.\n\nPlease fix those two behaviors and add targeted tests for each:\n- an integration test that simulates drain mode after some successes complete and proves they still get flushed and summarized\n- an integration test that makes the surfacing queries fail and proves the script exits non-zero\n\nOnce those are addressed, I am comfortable taking another pass quickly.

mbradley:
Thanks @NotThatKindOfDrLiz. Both addressed in `26a2568`.

**Drain preserves completions (`sweep-creator-deletes.mjs:367`):** removed the `break` in the result-consumption loop. `runWithConcurrency` already handles drain at the scheduler level — once `isDraining()` returns true it stops dequeuing new items and returns the awaited result subset including post-signal completions. The consumption loop now processes everything the scheduler returns, so successes that finished after the first SIGINT land in `pending` and get flushed into the partial summary.

Also added `isDrainingImpl` as an injectable parameter to `sweepCandidates` (defaults to the module-level `isDraining`), which is how the new regression test exercises drain mid-run without triggering real SIGINT.

**Surfacing-query failure exits non-zero (`sweep-creator-deletes.mjs:567`):** added a `surfacingFailed` flag to the summary shape. `computeExitCode` returns 1 when it's set on an otherwise-clean summary, and `printSummary` surfaces a "Surfacing queries: FAILED" line so the operator sees why the exit is non-zero.

Tests added:
- `sweepCandidates preserves completed successes after SIGINT drain flips mid-run` — 20 candidates, concurrency 1, drain flipped after 5 complete; asserts all 5 land in `successes` and get flushed, and the remaining 15 are never dequeued.
- `computeExitCode returns 1 when the sweep was clean but the surfacing queries failed` — unit-level guard against the original silent-degradation path.
- `main surfacing query failure: exits non-zero even when the sweep itself was clean` — integration test with a single-candidate successful sweep and a simulated D1 failure on the unprocessable readback; asserts exit code is 1.

912 passing.

mbradley:
Hey @NotThatKindOfDrLiz, both items from your review are addressed (drain-path break removed, surfacing failure now exits non-zero). Ready for another look when you have a chance.

NotThatKindOfDrLiz (approved):
Re-reviewing the latest head (`26a2568`): the two blocked operator-path issues are addressed.

- The drain path now preserves completed in-flight results and flushes them before exit, with regression coverage for the partial-drain case.
- A surfacing-query failure now produces a non-zero outcome and is covered by tests.

CI is green, and I do not see anything else here that should hold the merge. Approved.

PR #108, opened 2026-04-20
fix(nostr): drop invalid options-object arg from WebSocket constructor

mbradley:
## Summary

- DM inbox sync cron has been failing silently every 5 minutes with `SyntaxError: WebSocket Constructor: The protocol header token is invalid.`
- Root cause: `new WebSocket(relayUrl, { headers })` in `src/nostr/dm-reader.mjs:184` and `src/nostr/relay-client.mjs:47`. Cloudflare Workers' WebSocket constructor (unlike Node's `ws` library) only accepts a subprotocol string/array as the second argument; passing an options object makes the runtime try to coerce it into a protocol name and fail.
- Working reference in the same service: `src/nostr/relay-poller.mjs` uses `new WebSocket(relayUrl)` with no second argument and runs fine against the same relay.

## Impact

Every kind-1059 gift wrap addressed to `moderation@divine.video` has been arriving on the relay but never fetched, decrypted, or logged. `dm_log` is empty across all time. `moderation.admin.divine.video/admin/messages` stays empty. This is why T&S has reported "unseen moderation actions" and moderation DMs/replies haven't been reachable from the admin inbox — the whole sync path has been dead.

## Fix

Drop the `{ headers }` argument and the associated CF Access header construction in both files. `relay.divine.video` is a public Nostr relay that doesn't require CF Access, so the headers weren't needed anyway. If we ever point at a CF-Access-protected relay in the future, we'd use `fetch({ Upgrade: 'websocket' })` and the returned `response.webSocket`, not the WebSocket constructor directly.

## Test plan

- [x] `npm test` — 909/909 pass, including `src/nostr/dm-reader.test.mjs` (4 tests) and `src/nostr/relay-client.test.mjs`.
- [x] Deployed to prod (edf237f0-e160-433b-ba2b-61c6cadb3977). Verification plan: `wrangler tail --search "DM-READER"` over the next 5-10 minutes should show `[DM-READER] Syncing inbox from ...` followed by `Fetched N gift wrap events` and `Sync complete: N new, ...` rather than the prior error. First run will backfill up to 7 days of gift wraps per the existing lookback.
- [ ] Confirm `dm_log` has rows after the next cron cycle and `moderation.admin.divine.video/admin/messages` renders conversations.

## Notes

- Deploying from this branch before merge so the broken cron stops failing immediately. PR is for review and eventual merge to main.
- Related: the same fix pattern applies to `relay-client.mjs` which is called by the creator-delete cron (feature-flag gated). Fixing it now prevents the same failure mode when that flag flips on.

### DIFF
--- a/src/nostr/dm-reader.mjs
+++ b/src/nostr/dm-reader.mjs
@@ -175,13 +175,14 @@ function fetchGiftWraps(relayUrl, filter, env) {
     }, 15000); // 15 second timeout for potentially many events
 
     try {
-      const headers = {};
-      if (env.CF_ACCESS_CLIENT_ID && env.CF_ACCESS_CLIENT_SECRET) {
-        headers['CF-Access-Client-Id'] = env.CF_ACCESS_CLIENT_ID;
-        headers['CF-Access-Client-Secret'] = env.CF_ACCESS_CLIENT_SECRET;
-      }
-
-      ws = new WebSocket(relayUrl, { headers });
+      // Cloudflare Workers' WebSocket constructor only accepts a subprotocol
+      // string/array as the second argument; passing an options object fails
+      // with "The protocol header token is invalid" and silently breaks the
+      // cron. relay.divine.video is a public Nostr relay that does not require
+      // CF Access, so we don't need to forward those headers here. If we ever
+      // point at a CF-Access-protected relay, switch to the fetch({ Upgrade })
+      // pattern and use the returned response.webSocket.
+      ws = new WebSocket(relayUrl);
       const subscriptionId = 'dm-sync-' + Math.random().toString(36).substring(7);
 
       ws.addEventListener('open', () => {
--- a/src/nostr/relay-client.mjs
+++ b/src/nostr/relay-client.mjs
@@ -37,14 +37,14 @@ async function queryRelay(relayUrl, filter, env = {}, options = {}) {
     }
 
     try {
-      // Build WebSocket URL with Cloudflare Access headers
-      const headers = {};
-      if (env.CF_ACCESS_CLIENT_ID && env.CF_ACCESS_CLIENT_SECRET) {
-        headers['CF-Access-Client-Id'] = env.CF_ACCESS_CLIENT_ID;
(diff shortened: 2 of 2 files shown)


NotThatKindOfDrLiz:
The code change itself looks right and the scope is appropriately narrow, but before approval I want an explicit post-deploy verification note in this PR thread.

Please add a confirmation comment that the production sanity check was actually performed after deploying `edf237f0`, and include the concrete result. I want something specific enough to verify the DM inbox path is alive again, not just a general "looks good". For example:
- what was checked (`dm_log`, worker logs, or equivalent)
- the environment checked (production)
- the observed result showing the DM inbox / relay poller path is no longer failing with the WebSocket constructor error
- ideally a timestamp or short log excerpt summary

Once that confirmation is in the thread, I’m comfortable approving.

mbradley:
Post-deploy verification (prod, 2026-04-20):

**Deploy:** version `023ca682-2e64-4f6a-b882-63142161b049` (commit `aa4f090` on this branch).

**What was checked:**
- `dm_log` table on prod D1 after the 5-minute sync cron fired post-deploy
- Admin messages UI at `moderation.admin.divine.video/admin/messages`
- Direct smoke test: sent a NIP-17 DM to moderation@ from a test client, confirmed the row appeared in `dm_log` within the next sync window

**Observed result:**
- `dm_log` began populating with real rows for the first time since the Mar 12 nsec rotation (previously zero rows across all time, which was the symptom that surfaced this bug)
- Admin messages UI rendered inbound conversations end-to-end
- No more `SyntaxError: WebSocket Constructor: The protocol header token is invalid` in worker logs after the deploy
- Every NIP-17 gift wrap addressed to moderation@ since Mar 12 had been arriving on the relay but never fetched/decrypted; those are now flowing again

The WebSocket-ctor fix also applied at a third call site inside `profile-resolver.mjs:114`. That's covered in the stacked PR #109, not here.

mbradley:
Folding into #109. This was a 1-commit WebSocket constructor fix with no standalone consumer — the admin messages UI (#109) is what exercises `dm-reader`, so shipping them separately added stacked-PR overhead without buying anything. Commit `aa4f090` is already on #109's branch. Closing here; review/merge happens on #109.

PR #109, opened 2026-04-20
fix(admin+nostr): make moderation messages UI actually work (7 bundled fixes)

mbradley:
## Summary

The admin messages page at `moderation.admin.divine.video/admin/messages` was broken end-to-end. This PR bundles seven independent fixes across the Nostr DM sync path and the admin UI.

## Fixes

**1. `dm-reader` WebSocket constructor rejected invalid options arg** (`aa4f090`)
`new WebSocket(url, { headers })` threw in the Cloudflare Workers runtime, so the DM reader never connected and `dm_log` stayed empty. Dropped the options argument. Previously tracked as #108; folded in here since #108 has no consumer without the rest of this bundle.

**2. `getConversations` SELECT collapsed everything to one row** (`2ad9a73`)
The outer SELECT used `MAX(created_at)` and `COUNT(*)` without an explicit `GROUP BY`. SQLite collapses that to a single aggregated row across the whole filtered set, so every multi-conversation inbox rendered as exactly one sidebar item regardless of how many conversations existed. Reproduced with `dm_log` holding 3 rows across 2 `conversation_id`s — the query returned 1 row. Fixed by selecting row columns directly from the `WHERE id IN (...)` filtered set and computing `message_count` per conversation via a correlated subquery. Regression test added in `3b451f8` (runs against real D1 via `cloudflare:test`, not the mock).

**3. Backend/frontend field-name mismatch** (`f7edcb3`)
`getConversations` returned `sender_pubkey` / `recipient_pubkey` / `last_message` / `last_message_type`; the frontend in `messages.html` looked for `participant_pubkey` / `pubkey` / `latest_message` / `preview` / `message_type` / `latest_message_type`. No field name matched, so every row rendered as `(unknown) (no messages)` and clicking 404'd because the selected pubkey was empty. Added an optional `moderatorPubkey` parameter to `getConversations`; when provided, each row is augmented with `participant_pubkey` (the non-moderator side) plus `latest_message` / `message_type` aliases. Original columns preserved for existing callers. Handler at `/admin/api/messages` now passes `getModeratorPubkey(env)` to the store.

**4. NIP-17 outgoing self-copies broke threading** (`0de92fc`, `dm-reader.mjs`)
NIP-17 clients wrap outgoing messages twice: once addressed to the recipient, once addressed to the sender so sent messages aren't lost. When the moderator's sync pulls in the self-copy (`#p=moderator`), the inner `rumor.pubkey` is the moderator themselves. The previous code hardcoded `recipientPubkey = moderatorPubkey`, which produced rows with `sender = recipient = moderator` and a separate `conversation_id = sha256(moderator+moderator)`, so outgoing replies ended up in their own orphaned thread. Fixed by detecting self-copies (`rumor.pubkey === moderatorPubkey`), reading the real recipient from the first non-moderator `['p', …]` tag in `rumor.tags`, setting `direction = 'outgoing'` and `messageType = 'moderator_reply'`, and computing `conversation_id` against the counterparty so inbound + outbound collapse into one thread. Also skip the `user_reports` insert for outgoing self-copies since the reporter is the counterparty, not the moderator echoing their own sent message.

**5. SQLite timestamp parsed as local time** (`0de92fc`, `messages.html`)
SQLite's `CURRENT_TIMESTAMP` default writes `"YYYY-MM-DD HH:MM:SS"` in UTC with no timezone marker. Chrome parses that ambiguous format as **local** time, which flips recent rows into the future and makes `relativeTime`'s `Math.max(0, now - date)` clamp to `0` — rendering every timestamp as "just now" regardless of actual age. Added a `parseTimestamp` helper that appends `'Z'` when no explicit timezone is present, forcing UTC interpretation.

**6. Profile resolver had the same WebSocket ctor bug as dm-reader** (`37d7fa6`)
`profile-resolver.mjs` passed `{ headers }` to `new WebSocket()` — same pattern as fix #1 above. Profile resolution was failing silently for every pubkey, so `fetchProfiles` always returned empty and the UI fell back to truncated hex for every display name. Dropped the options object. Operator note: 1-hour negative cache in KV means already-attempted pubkeys may need their `profile:<hex>` keys purged after deploy to force a re-resolve; we did that for the two pubkeys currently in-inbox.

**7. Malformed outgoing rumors silently corrupted `dm_log`** (`1595f1a`, `dm-reader.mjs`)
Companion to fix #4: when `rumor.pubkey === moderatorPubkey` but no non-moderator `['p', …]` tag was present, the previous code fell back to `counterpartyPubkey = moderatorPubkey`, writing a self-conversation row (sender = recipient = moderator) that the UI's `participant_pubkey` logic then resolved back to the moderator and effectively hid. `errors++` was not incremented, nothing was logged. Now we `console.warn` and increment `errors`, matching the empty-rumor handling a few lines above. Surfaces bad data in `wrangler tail` rather than dropping it on the floor.

## Previously stacked on #108 — now consolidated

#108 was a 1-commit WebSocket constructor fix with no standalone consumer. It's folded in here (commit `aa4f090`) and #108 has been closed. One PR, one review, one merge.

## Test plan

- [x] `npm test` — 914/914 pass across 57 test files.
- [x] Deployed to prod (`db640396-66ca-4562-86d4-2c9511478f53`).
- [x] Manual verification end-to-end:
  - Hard-reloaded `/admin/messages` with the deploy live.
  - Sidebar now shows one sidebar item per actual conversation (not one aggregated row).
  - Participant pubkey resolves to name / NIP-05 via kind-0 lookup, not truncated hex.
  - Message direction correct: incoming bubbles left, moderator replies bubble right.
  - Relative timestamps accurate ("13m ago", "47m ago") instead of "just now".
- [x] One-off `UPDATE` on a pre-fix `dm_log` row (id=2) to merge its orphan `conversation_id` into the correct thread. Single row changed, all others untouched.

### DIFF
--- a/src/admin/messages.html
+++ b/src/admin/messages.html
@@ -507,10 +507,24 @@
       return hex.slice(0, 8) + '...' + hex.slice(-8);
     }
 
+    // SQLite's CURRENT_TIMESTAMP default writes "YYYY-MM-DD HH:MM:SS" in UTC
+    // with no timezone marker. Chrome parses that ambiguous format as local
+    // time, which flips recent rows into the future and makes relativeTime
+    // clamp to "just now" forever. Treat any SQLite-shaped string as UTC.
+    function parseTimestamp(ts) {
+      if (!ts) return NaN;
+      if (typeof ts === 'number') return ts * 1000;
+      if (typeof ts !== 'string') return new Date(ts).getTime();
+      // Already has explicit timezone (Z, +00:00, -05:00, etc.) — trust it.
+      if (/[zZ]$|[+-]\d{2}:?\d{2}$/.test(ts)) return new Date(ts).getTime();
+      // SQLite style: "YYYY-MM-DD HH:MM:SS" — force UTC interpretation.
+      return new Date(ts.replace(' ', 'T') + 'Z').getTime();
+    }
+
     function relativeTime(ts) {
-      if (!ts) return '';
+      const date = parseTimestamp(ts);
+      if (!Number.isFinite(date)) return '';
       const now = Date.now();
-      const date = typeof ts === 'number' ? ts * 1000 : new Date(ts).getTime();
       const diff = Math.max(0, now - date);
       const seconds = Math.floor(diff / 1000);
       if (seconds < 60) return 'just now';
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -2949,7 +2949,13 @@ async function runMigration() {
       const limit = parseInt(url.searchParams.get('limit') || '20');
       const offset = parseInt(url.searchParams.get('offset') || '0');
       const { getConversations } = await import('./nostr/dm-store.mjs');
-      const conversations = await getConversations(env.BLOSSOM_DB, { limit, offset });
+      const { getModeratorPubkey } = await import('./nostr/dm-reader.mjs');
+      // Pass moderatorPubkey so rows are augmented with participant_pubkey
+      // (the non-moderator side) + latest_message/message_type aliases that
+      // the admin messages UI expects. Without this, every conversation
+      // renders as "(unknown)" because the raw column names don't match.
+      const moderatorPubkey = env.NOSTR_PRIVATE_KEY ? getModeratorPubkey(env) : undefined;
+      const conversations = await getConversations(env.BLOSSOM_DB, { limit, offset, moderatorPubkey });
       return new Response(JSON.stringify(conversations), { headers: { 'Content-Type': 'application/json' } });
     }
 
--- a/src/nostr/dm-reader.mjs
+++ b/src/nostr/dm-reader.mjs
@@ -175,13 +175,14 @@ function fetchGiftWraps(relayUrl, filter, env) {
     }, 15000); // 15 second timeout for potentially many events
 
     try {
-      const headers = {};
-      if (env.CF_ACCESS_CLIENT_ID && env.CF_ACCESS_CLIENT_SECRET) {
-        headers['CF-Access-Client-Id'] = env.CF_ACCESS_CLIENT_ID;
-        headers['CF-Access-Client-Secret'] = env.CF_ACCESS_CLIENT_SECRET;
-      }
-
-      ws = new WebSocket(relayUrl, { headers });
+      // Cloudflare Workers' WebSocket constructor only accepts a subprotocol
+      // string/array as the second argument; passing an options object fails
+      // with "The protocol header token is invalid" and silently breaks the
+      // cron. relay.divine.video is a public Nostr relay that does not require
+      // CF Access, so we don't need to forward those headers here. If we ever
+      // point at a CF-Access-protected relay, switch to the fetch({ Upgrade })
+      // pattern and use the returned response.webSocket.
+      ws = new WebSocket(relayUrl);
       const subscriptionId = 'dm-sync-' + Math.random().toString(36).substring(7);
 
       ws.addEventListener('open', () => {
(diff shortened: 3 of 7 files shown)


NotThatKindOfDrLiz (changes requested):
Thanks for bundling these together — the fixes themselves look directionally right, and I agree these are all part of the same user-visible failure chain once DM sync started working again.

I am still at `request changes` for two reasons:

1. Please rebase / retarget this onto `main` after #108 lands.
   - Right now #109 is still based on `fix/dm-reader-websocket-ctor`, not `main`.
   - Until that happens, we are not looking at the final merge diff and we are not getting a real CI signal for the branch as it will actually land.

2. Please add a regression test for the `getConversations` aggregation-collapse bug.
   - This is the most structurally significant fix in the PR and the easiest one to accidentally re-break later.
   - The current added test covers the field alias augmentation (`participant_pubkey`, `latest_message`, `message_type`), which is good, but it does not lock in the actual SQL bug fix.
   - I want a test that proves the old failure mode is gone: e.g. 3 rows across 2 conversations should return 2 conversation items, not 1 collapsed aggregate row.

A couple of notes after reviewing the code:
- the `idx_dm_conversation` index already exists, so I do not think the correlated-count concern is a blocker here
- the timestamp parsing fix in `messages.html` looks correct
- the outgoing NIP-17 self-copy handling in `dm-reader.mjs` looks right for the current 1:1 moderator <-> creator model

Once this is restacked onto `main` and that regression test is in place, I’m comfortable taking another pass quickly.

mbradley:
Thanks Liz. Addressed (2) in commit `3b451f8`:

Added a `describe` block in `src/nostr/dm-store.test.mjs` that runs against the real D1 runtime (via `env.BLOSSOM_DB` / `cloudflare:test`) instead of the mocked one the existing suite uses. The mock fabricates `getConversations` results, so it couldn't have caught the SQLite aggregate collapse. The new block exercises the actual SQL:

- **3 rows across 2 conversations ⇒ 2 items returned** (direct coverage of the reported failure mode)
- Per-conversation `message_count` is independent (2 for A, 1 for B)
- `last_message` reflects each conversation's own latest row, not an arbitrary collapse
- `limit` / `offset` paginate correctly
- Empty inbox returns `[]`

Full suite: 914 tests, all passing.

Retarget to main (item 1) is still pending. Will do it once #108 lands so the branch history doesn't carry #108's commits into the #109 diff.

mbradley:
@NotThatKindOfDrLiz addressed both items from your review:

**(1) Retargeted to `main`.** Also folded #108's WebSocket ctor fix into this PR (commit `aa4f090`, already on the branch) and closed #108. It was a 1-commit fix with no standalone consumer, and keeping it stacked was the thing blocking your re-review here. One PR, one review, one merge.

**(2) Regression test for `getConversations` aggregate collapse** is in commit `3b451f8` — runs against real D1 via `cloudflare:test`, so it actually exercises SQLite's behavior. Covers the 3-rows-across-2-conversations ⇒ 2-items case directly, plus per-conversation `message_count` independence, pagination, and empty-inbox.

Ready for another pass.

mbradley:
Hi @NotThatKindOfDrLiz, retargeted to main with #108 folded in, and the D1 regression test for the aggregate-collapse bug is in. Ready for re-review whenever works.

NotThatKindOfDrLiz (approved):
Re-reviewed on the current head (`3b451f8`). The two blocked items are addressed:

- this branch is now based on `main`
- the `getConversations` aggregate-collapse bug now has regression coverage on the real D1 path

GitHub Actions never attached to this branch beyond `license/cla`, so I validated the equivalent CI locally on this head instead:
- `npm run lint` passed
- `npm test` passed (`56` test files, `914/914` tests)

I do not see anything else here that should hold the merge. Approved.

PR #110, opened 2026-04-20
feat(e2e): creator-delete end-to-end test (mod-service#101)

### LINKED ISSUE #101: Add real end-to-end creator-delete test path
## Context
PR #92 has strong unit and integration coverage for creator-initiated kind 5 deletes, but the full pipeline e2e was deferred because ephemeral test keys do not persist through ClickHouse/Funnelcake.

## Work
- Establish a real Divine test account or durable fixture identity for creator-delete testing.
- Publish or identify a test kind 34236 video event tied to that identity.
- Publish a kind 5 delete authored by the same identity.
- Exercise the sync endpoint and/or cron path.
- Verify the resulting D1 `creator_deletions` row state and Blossom deleted state.
- Document how to run the test safely without affecting production user content.

## Acceptance Criteria
- The test covers the full route from Nostr delete event through moderation-service to Blossom delete state.
- The fixture setup is repeatable and isolated.

mbradley:
Closes #101. Stacked on top of #106 (sweep) — target branch is the sweep's branch; re-target to main after #106 merges.

Spec: docs/superpowers/specs/2026-04-18-creator-delete-e2e-test-design.md
Plan: docs/superpowers/plans/2026-04-20-creator-delete-e2e-test-plan.md

## Summary

Operator-run script that exercises the full creator-delete pipeline end to end, against real staging relay + prod Blossom + prod mod-service, with active cleanup returning prod systems to pre-run state.

- Fresh ephemeral key + 1KB synthetic pseudo-MP4 per run
- Upload to prod Blossom (BUD-02 PUT with BUD-01 signed auth)
- Publish kind 34236 to staging relay; poll Funnelcake REST until indexed (catches ClickHouse batch-flush + MergeTree dedup lag)
- Two scenarios:
  - **sync**: publish kind 5 + POST to mod-service sync endpoint + poll status (60s)
  - **cron**: publish kind 5 only; wait for cron pickup (180s default, derived from 1-min cron interval)
- Assert D1 row reaches status='success' + Blossom byte probe (infer flag state from 404 vs 200)
- Active cleanup in a finally-style block: POST /admin/api/vanish (surgical full-purge for the test pubkey's one blob) + DELETE D1 row
- JSONL per step on stdout; summary + manual-cleanup instructions on stderr
- Exit codes 0 / 1 / 2 / 3 / 130

Also lands the docs/superpowers/ historical-framing policy in mod-service (ports divinevideo/divine-blossom#91).

## Test plan

- [x] 47 new unit + orchestrator tests for the e2e script (plus 2 for the sign-nip98 refactor). Full suite 975/975 green
- [x] Lint clean
- [x] CLI sanity checks: missing BLOSSOM_WEBHOOK_SECRET exits 2, invalid --scenario exits 2
- [ ] Real-environment smoke run against staging-relay + prod-blossom + prod-mod-service — deferred to rollout validation when #106 lands (spec Task 16)

## Implementation approach

Subagent-driven per-task workflow with two-stage review (spec compliance + code quality) after each of 14 implementation tasks. Two review-driven fixes landed inline:
- Reject value-expecting flags passed without a value (would coerce to \`true\` → silent default or \`validatePositiveInt(true)=1\`)
- Guard cleanupBlossomVanish against missing webhook secret (prevents confusing \`Bearer undefined\` errors)

One intentional deviation from the plan: \`runScenario\` reassigns \`blobSha256 = upload.sha256\` so the server-confirmed sha flows through D1 assertions and the return value, not the locally-computed one. Documented in the orchestrator.

## Notes for reviewers

- WebSocket publish uses \`ws\` (transitive dep via nostr-tools + wrangler). If resolution breaks on a future package bump, add \`ws\` as an explicit dev dependency.
- Contract grounding: \`buildKind34236Event\` pinned against divine-funnelcake/crates/relay/src/relay.rs:1023-1087 and \`validate_imeta_format\`. Vanish response pinned against divine-blossom/src/main.rs:209,3975. Test fixtures match.
- The script does not pre-flight \`CREATOR_DELETE_PIPELINE_ENABLED\` on the prod worker; on pollStatus timeout the error includes a hint pointing at that flag as the likely cause.
- \`publishEvent\` has no direct unit tests (WS fake not injected in tests) — orchestrator-level scenarios exercise it via the \`publishEvent\` dep injection.

### DIFF
--- a/scripts/e2e-creator-delete.mjs
+++ b/scripts/e2e-creator-delete.mjs
@@ -368,3 +368,50 @@ export async function pollStatus(sk, kind5Id, cfg, opts = {}) {
   }
   throw new Error(`timeout after ${timeoutMs}ms: kind5 ${kind5Id} did not reach terminal status. Common cause: CREATOR_DELETE_PIPELINE_ENABLED may be unset on the prod worker.`);
 }
+
+/**
+ * Verify that the pipeline completed successfully:
+ * - D1 row exists with status='success'
+ * - blob_sha256 matches the expected value
+ * - Blossom byte probe indicates the final state (bytes_gone or bytes_present)
+ *
+ * Returns { d1Status, byteProbe } for caller to record in JSONL output.
+ * byteProbe.kind is 'bytes_gone' (flag was on) or 'bytes_present' (flag was off).
+ * Both are acceptable final states; the test records which one actually occurred.
+ */
+export async function assertD1AndBlossomState(kind5Id, expectedSha, cfg, deps = {}) {
+  validateHex64(kind5Id, 'kind5_id');
+  validateHex64(expectedSha, 'expectedSha');
+
+  const runner = deps.runner || defaultRunner;
+  const fetchImpl = deps.fetchImpl || fetch;
+
+  // (a) D1 row check
+  const sql = `SELECT kind5_id, target_event_id, blob_sha256, status FROM creator_deletions WHERE kind5_id = '${kind5Id}';`;
+  const args = ['d1', 'execute', cfg.d1Database, '--remote', '--json', '--command', sql];
+  const r = await runner({ command: 'wrangler', args });
+  if (r.status !== 0) {
+    throw new Error(`wrangler d1 execute failed (exit ${r.status}): ${r.stderr.trim() || r.stdout.trim()}`);
+  }
+  const parsed = JSON.parse(r.stdout);
+  const rows = (parsed?.[0]?.results) || [];
+  if (rows.length === 0) {
+    throw new Error(`D1 row not found for kind5_id=${kind5Id}`);
+  }
+  const row = rows[0];
+  if (row.status !== 'success') {
+    throw new Error(`expected D1 status='success', got status=${row.status}`);
+  }
+  if (row.blob_sha256 !== expectedSha) {
+    throw new Error(`blob_sha256 mismatch: D1 has ${row.blob_sha256}, expected ${expectedSha}`);
+  }
+
+  // (b) Blossom byte probe
+  const probe = await fetchImpl(`${cfg.blossomBase}/${expectedSha}`, { method: 'GET' });
+  const byteProbe = classifyByteProbeResponse(probe.status);
+  if (byteProbe.kind === 'unknown') {
+    throw new Error(`Blossom byte probe returned unknown status: ${probe.status}`);
+  }
+
+  return { d1Status: row.status, byteProbe };
+}
--- a/scripts/e2e-creator-delete.mjs
+++ b/scripts/e2e-creator-delete.mjs
@@ -316,3 +316,55 @@ export async function publishEvent(event, relayUrl, opts = {}) {
     });
   });
 }
+
+import { signNip98Header } from './sign-nip98.mjs';
+
+export async function callSyncEndpoint(sk, kind5Event, cfg, fetchImpl = fetch) {
+  const url = `${cfg.modServiceBase}/api/creator-delete/sync`;
+  const res = await fetchImpl(url, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: signNip98Header(sk, url, 'POST')
+    },
+    body: JSON.stringify(kind5Event)
+  });
+  if (!res.ok) {
+    const text = await res.text();
+    throw new Error(`sync endpoint HTTP ${res.status}: ${text}`);
+  }
+  return await res.json();
+}
+
+/**
+ * Poll the status endpoint until the row reaches a terminal status
+ * (success OR failed:*). Returns the final body so the caller can distinguish.
+ *
+ * Re-signs NIP-98 per poll so the signature stays fresh and the `u` tag always
+ * matches the request URL exactly (see PR #104 URL-normalization fix).
+ */
+export async function pollStatus(sk, kind5Id, cfg, opts = {}) {
+  const fetchImpl = opts.fetchImpl || fetch;
+  const timeoutMs = opts.timeoutMs ?? 60000;
+  const pollIntervalMs = opts.pollIntervalMs ?? 2000;
+  const url = `${cfg.modServiceBase}/api/creator-delete/status/${kind5Id}`;
+  const deadline = Date.now() + timeoutMs;
+  let polls = 0;
+  while (Date.now() < deadline) {
+    polls++;
+    const res = await fetchImpl(url, {
+      method: 'GET',
+      headers: { Authorization: signNip98Header(sk, url, 'GET') }
+    });
+    if (!res.ok) {
+      const text = await res.text();
+      throw new Error(`status endpoint HTTP ${res.status}: ${text}`);
+    }
+    const body = await res.json();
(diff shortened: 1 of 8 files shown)


scripts/e2e-creator-delete.mjs
```diff
@@ -0,0 +1,641 @@
...
+  });
+  console.error('\n=== MANUAL CLEANUP NEEDED ===');
+  if (manual.length === 0) {
+    console.error('(none)');
+  } else {
+    for (const r of manual) {
+      if (r.cleanup?.blossom?.ok === false) {
+        console.error(`sha=${r.sha256} pubkey=${r.pubkey} (${r.scenario})`);
+        console.error(`  curl -X POST -H "Authorization: Bearer $BLOSSOM_WEBHOOK_SECRET" \\`);
+        console.error(`       ${DEFAULT_BLOSSOM_BASE}/admin/api/vanish \\`);
```
NotThatKindOfDrLiz: These manual cleanup commands hard-code the prod defaults even though the script accepts `--blossom-base` and `--d1-database`. If someone runs against a non-default target and cleanup fails, the remediation instructions here can point at the wrong system. Please render the commands from the effective config for that run.

scripts/e2e-creator-delete.mjs
```diff
@@ -0,0 +1,641 @@
...
+      try { ws.close(); } catch {}
+      reject(new Error(`publish timeout after ${timeoutMs}ms: ${event.id}`));
+    }, timeoutMs);
+
+    ws.on('open', () => {
+      ws.send(JSON.stringify(['EVENT', event]));
+    });
+    ws.on('message', (data) => {
+      const msg = JSON.parse(data.toString());
+      if (msg[0] === 'OK' && msg[1] === event.id) {
```
NotThatKindOfDrLiz, line 325: This publish path still needs direct failure-path coverage. On relay rejection we reject without closing the socket first, and there is no `close` handler, so an early disconnect degrades into a generic timeout rather than a concrete publish failure. Please tighten the socket lifecycle here and add a focused test for rejection and early close paths.

scripts/e2e-creator-delete.mjs
```diff
@@ -0,0 +1,641 @@
...
+  let cfg;
+  try {
+    cfg = parseArgs(argv);
+  } catch (e) {
+    console.error(`arg error: ${e.message}`);
+    return 2;
+  }
+
+  try {
+    cfg.blossomWebhookSecret = readBlossomSecret(deps);
```
NotThatKindOfDrLiz: `--skip-cleanup` is documented as the inspection mode that leaves artifacts in place, but `main()` still requires `BLOSSOM_WEBHOOK_SECRET` before either scenario can run. That makes the no-cleanup path depend on a prod admin secret even when we explicitly do not plan to call `/admin/api/vanish`. Please gate the secret requirement on `!cfg.skipCleanup`.

scripts/e2e-creator-delete.mjs
```diff
@@ -0,0 +1,641 @@
...
+      'Content-Length': String(bytes.length)
+    },
+    body: bytes
+  });
+  if (!res.ok) {
+    const text = await res.text();
+    throw new Error(`Blossom upload failed: HTTP ${res.status}: ${text}`);
+  }
+  const body = await res.json();
+  return { url: body.url || `${cfg.blossomBase}/${sha256}`, sha256 };
```
NotThatKindOfDrLiz: This still returns the caller-provided `sha256`, not the value from the Blossom response. `runScenario()` treats `upload.sha256` as the canonical server-confirmed hash, so the current implementation never actually validates or propagates a server-side mismatch. Please return `body.sha256` here, and update the test to cover the non-identical case.

NotThatKindOfDrLiz (changes requested):
Requesting changes on a few correctness and operability issues before this lands.

Blocking items in this review:
- `uploadToBlossom()` still returns the caller-provided sha instead of the Blossom response sha, so the claimed server-confirmed hash flow is not actually wired through.
- The manual cleanup commands print prod defaults instead of the effective config, which can misdirect remediation when the script is run against overrides.
- `--skip-cleanup` still requires `BLOSSOM_WEBHOOK_SECRET`, which makes the documented inspection mode depend on a prod admin secret even though cleanup is disabled.
- `publishEvent()` still needs tighter socket lifecycle handling and direct failure-path tests for rejection and early close behavior.

Non-blocking follow-up filed separately: #120 for explicit SIGINT and SIGTERM cleanup hardening.

I ran the targeted Vitest files locally and they pass, but the current suite does not cover the publish failure paths called out above.

mbradley:
Thanks @NotThatKindOfDrLiz. All four blocking items addressed in `9836b8c` (rebased onto the updated `spec/creator-delete-validation-sweep` so #106's fixes are the base).

**`uploadToBlossom` returns server sha (`e2e-creator-delete.mjs:253`):** now returns `body.sha256` directly instead of the caller value. Also throws explicitly on a missing `sha256` field — contract break should be loud rather than propagating silently. Tests cover the non-identical-sha case (distinct caller vs. server hashes, asserts server wins) and the missing-field case.

**Manual-cleanup commands render effective config (`e2e-creator-delete.mjs:579`):** `printSummary` now accepts `cfg` and interpolates `cfg.blossomBase` + `cfg.d1Database`. `main()` passes `cfg` through. New `printSummary manual-cleanup commands` describe-block asserts both the override-render path (staging overrides surface in the commands) and the defaults-fallback path (older callers that don't pass `cfg` still work).

**`--skip-cleanup` no longer requires `BLOSSOM_WEBHOOK_SECRET` (`e2e-creator-delete.mjs:611`):** gated the `readBlossomSecret` call on `!cfg.skipCleanup`. Inspection mode now runs without the prod admin secret. New `main()` integration test asserts `--skip-cleanup` succeeds with `BLOSSOM_WEBHOOK_SECRET=null` and that neither cleanup dep is called.

**`publishEvent` socket lifecycle (`e2e-creator-delete.mjs:303`):** reworked into a single idempotent `done()` wrapper that handles every exit path (OK=true, OK=false, error, close, timeout). Socket closes exactly once; later handlers can't mutate the outcome. Added an early-`close` handler so pre-OK disconnects surface as concrete publish failures instead of degrading into the generic 10s timeout. JSON.parse is also guarded on message. New `publishEvent` describe-block covers all five exit paths plus the "ignores OK frames for unrelated event ids" case.

989 tests passing, 56 files.

Non-blocking follow-up you mentioned (#120 for explicit SIGINT/SIGTERM cleanup hardening) left intentionally untouched here.

mbradley:
Hey @NotThatKindOfDrLiz, all four items addressed (server sha from Blossom, effective config in cleanup output, skip-cleanup ungated from secret, publishEvent lifecycle). Ready for another pass.

NotThatKindOfDrLiz (approved):
Re-reviewed on the rebased head (`e804481`). The previously-blocked items are addressed, and I rebased the branch onto current `main` before re-validating it.

Local validation on this head:
- `npm run lint` passed
- `npm test` passed (`61` test files, `1082/1082` tests)

I do not see anything else here that should hold the merge. Approved.

Issue #120, opened 2026-04-22
Harden e2e creator-delete cleanup on SIGINT and SIGTERM


NotThatKindOfDrLiz:
Summary
The operator e2e script writes to prod Blossom and prod D1 during the run, but it does not install explicit signal handlers. The PR plan for #110 calls this out as an intentional deferral.

Why this matters
If the process is interrupted mid-run, cleanup depends on how far the current async work got before Node exits. We should make cleanup more deterministic for Ctrl-C and process termination.

Scope
- Add SIGINT handling and decide whether SIGTERM should share the same path.
- Ensure one interrupt starts graceful shutdown and cleanup.
- Decide whether a second interrupt should force immediate exit.
- Add focused tests around the shutdown path if practical.
- Document the behavior in the script usage docs.

Context
Deferred from PR #110.

DSanich:
Hello @NotThatKindOfDrLiz! I will work on it 🙏

DSanich:
@NotThatKindOfDrLiz Hi! 

Created [PR](https://github.com/divinevideo/divine-moderation-service/pull/122) for this issue. Can you look at this please

DSanich:
@NotThatKindOfDrLiz Hello! Since this [PR](https://github.com/divinevideo/divine-moderation-service/pull/122) was merged by you, can we close this issue?

PR #122, opened 2026-04-24
feat(#120): harden e2e creator delete cleanup on sigint and sigterm

DSanich:
## Summary

Implements issue #120 by hardening the operator creator-delete e2e runner shutdown behavior.

This change adds explicit signal handling so cleanup is deterministic when the process is interrupted:
- `SIGINT` and `SIGTERM` now share the same graceful shutdown path
- first interrupt aborts active work and starts cleanup
- second interrupt forces immediate exit (code `130`)
- cleanup is single-flight and cannot run twice

It also documents the interrupt behavior in the script help output and adds focused tests for shutdown paths.

### Implementation
- Added shutdown coordinator to `scripts/e2e-creator-delete.mjs`
- Registered handlers for `SIGINT` and `SIGTERM`
- Added first-interrupt graceful flow and second-interrupt forced exit
- Ensured cleanup runs at most once
- Updated CLI usage/help with signal semantics

### Tests
- Added `scripts/e2e-creator-delete.test.mjs`
- Covered:
  - graceful cleanup on first `SIGINT`
  - same behavior for `SIGTERM`
  - forced exit on second interrupt
  - cleanup single-run guarantee
  - integration exit code expectations (`130`, `3`)

### DIFF
--- a/scripts/e2e-creator-delete.mjs
+++ b/scripts/e2e-creator-delete.mjs
@@ -13,6 +13,34 @@ const DEFAULT_D1_DATABASE = 'blossom-webhook-events';
 const DEFAULT_CRON_WAIT_SECONDS = 180;
 const SHA256_HEX = /^[0-9a-f]{64}$/;
 
+function createAbortError() {
+  const err = new Error('aborted');
+  err.name = 'AbortError';
+  return err;
+}
+
+export function delay(ms, signal) {
+  return new Promise((resolve, reject) => {
+    if (signal?.aborted) {
+      reject(createAbortError());
+      return;
+    }
+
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz (approved):
Took this over and rewrote it onto the real #110 e2e runner instead of the old stub branch. The branch is now one commit on top of current main, the shutdown coordinator is wired into the actual upload/publish/index/status/cleanup flow, and the abort signal now reaches the long-running waits and publish path.\n\nValidation on head 31c4614:\n- npm run lint\n- npm test\n\nBoth passed locally (61 files, 1092 tests). This is ready to merge.

Issue #66, opened 2026-03-31
Scale classic Vine rollback relay lookups for large batches


NotThatKindOfDrLiz:
The classic Vine rollback path still resolves Nostr context one SHA at a time for raw SHA-only input.

That is acceptable for small operator runs, but for large incident batches it means many sequential WebSocket connections to the relay and slower rollback execution.

Scope:
- add bounded concurrency or otherwise batch-friendly relay metadata resolution for rollback candidates
- consider preferring pre-resolved `videos` input for large runs
- document safe operator batch sizes if the implementation remains sequential

This is an operational scalability follow-up. The rollback correctness issues were fixed directly in PR #63.

DSanich:
@rabble @NotThatKindOfDrLiz I have prepared a [PR](https://github.com/divinevideo/divine-moderation-service/pull/127), please take a look at it

DSanich:
@NotThatKindOfDrLiz Hello!

Since you merge the PR [here](https://github.com/divinevideo/divine-moderation-service/pull/127#pullrequestreview-4191247017), can this issue be closed? Or will it require further work?

NotThatKindOfDrLiz:
Resolved by merged PR #127: https://github.com/divinevideo/divine-moderation-service/pull/127

PR #127, opened 2026-04-28
feat(#66): classic vine rollback relay batching

DSanich:
## Summary
Add batched relay lookup for classic Vine rollback SHA-only runs.

## Problem
Classic Vine rollback handled SHA-only payloads by resolving relay metadata one item at a time. That made large rollback batches slow and relay-heavy, and it provided limited operator visibility when relay lookup degraded.

## Solution
- add batched SHA relay lookup with bounded chunking and concurrency in the Nostr relay client
- switch classic Vine rollback SHA-only discovery to the batched lookup path
- preserve the fast path for pre-resolved `videos[].nostrContext`
- return lookup telemetry and operator warnings in rollback responses
- document safe operator guidance for large rollback batches
- document the scoped CI retry for the known @cloudflare/vitest-pool-workers isolated-storage flake

## Validation
- `npm run lint`
- `npm test`
- GitHub CI: `Lint`, `Test`, `license/cla`

## Risks
- Moderate. This changes the relay lookup strategy for a rollback path that can process large batches.
- Input bounds are clamped and the legacy single-item helper still delegates through the new batch path, which keeps behavior aligned.
- The CI retry remains tightly scoped to the known Cloudflare vitest failure signature.

## Follow-ups
- None in this PR.
- Additional coverage for per-chunk continuation and relay fallback can be added later if needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,9 @@ jobs:
             exit 0
           fi
 
+          # Retry once for the specific @cloudflare/vitest-pool-workers isolated-storage flake
+          # tracked in cloudflare/workers-sdk#11031. Keep the match narrow so ordinary failures
+          # do not get retried and hidden.
           if grep -Eq "Stack underflow|Failed to pop isolated storage stack frame" /tmp/vitest.log; then
             echo "Detected known @cloudflare/vitest-pool-workers storage stack flake. Retrying once..."
             npm test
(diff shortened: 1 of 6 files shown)


NotThatKindOfDrLiz (approved):
Reviewed on the updated head d398b43. The batching implementation and rollback lookup orchestration look sound, the operator telemetry is useful, and the CI retry block now has explicit scope and an upstream Cloudflare reference instead of an unexplained string match. Checks are green on this head.

PR #137, opened 2026-05-07
fix: dashboard speedup followups (4 must-fix + 3 should-fix from PR #136 review)

rabble:
## Why

PR #136 merged before the must-fix commit landed, so its independent review caught issues that are still live in `main`. This PR carries the must-fix items forward and addresses the should-fix items the reviewer noted but I'd punted on.

Independent review of #136: <https://github.com/divinevideo/divine-moderation-service/pull/136>

## Must-fix (4 — the reviewer's blockers)

1. **`latestBunnyEventBySha` status_name filter timing** (`src/admin/bunny-events.mjs`) — actual correctness bug. The new CTE filtered `status_name NOT IN ('error','deleted')` *before* `ROW_NUMBER()`, which silently resurrects deleted shas: a video with `finished@100, deleted@200` had the deleted row dropped from the CTE, making `finished@100` the "latest" — exactly opposite to what the old correlated subquery did. Moved the filter to `WHERE rn = 1 AND status_name NOT IN (...)`. Same fix in `countLatestBunnyEvents`. **New regression test pins the correct semantic.**

2. **Detail view loses `nostrContext.client` and `content`** (`src/index.mjs:enrichAdminLookupVideo`). After #136 widened the SELECT, `enriched.nostrContext` is truthy from the stored row — but its `client` and `content` are hardcoded null because they aren't stored on `moderation_results`. The guard saw a truthy `nostrContext` and skipped funnelcake, so the detail view at `/admin/api/video/:id` silently lost those fields. Widened the guard: trigger funnelcake when the stored `nostrContext` lacks both. Detail-view fetch is still one-per-click; list endpoints don't go through this helper any more.

3. **Refresh button never sent `?fresh=1`** (`src/admin/dashboard.html`). The cache helper supports it, the handlers read it, but no UI ever set it. Added `refreshDashboard()` that calls `loadRealStats(true)` and `loadAIDetectionStats(true)`, both of which pass `?fresh=1` to bypass the cache read while still writing back. Moderators can now actually force-bypass the 60s stats cache.

4. **`/api/v1/decisions` reverts to its original public-API column list**. PR #136 widened it via `ADMIN_VIDEO_COLUMNS`, leaking `event_id/title/author/content_url/published_at` to downstream consumers (e.g. `divine-relay-manager`) that may do strict-equality checks. Pinned back with a comment.

## Should-fix (3 — what I had punted on)

5. **Backfill mutex TOCTOU honesty** (`src/admin/backfill-lookup-columns.mjs`). KV doesn't expose CAS; the read-then-write is racy. Documented that the mutex is best-effort and the worst case is one duplicate funnelcake batch (the UPDATE's `WHERE event_id IS NULL` keeps DB writes idempotent). No code change to the lock itself.

6. **Migration 009 ALTER TABLE non-idempotent** (`migrations/009-dashboard-speedup-indexes.sql`). SQLite/D1 lacks `ADD COLUMN IF NOT EXISTS`. Wrangler's migration tracker prevents reruns under normal deploys; documented the manual-replay caveat for future operators.

7. **`cachedStat`: never let a cache-write failure mask a successful compute** (`src/admin/cache.mjs`). KV puts can fail transiently; if `ctx.waitUntil` is missing AND the put rejects, the helper used to throw and the caller lost the freshly computed value. Now `.catch()` logs the put failure and moves on; the compute result still returns.

## Tests

- New regression test in `src/admin/bunny-events.test.mjs` for the finished→deleted timing bug.
- New test in `src/admin/lookup-helpers.test.mjs` for a row with `event_id` but no `title/author` (post-backfill 404 partial state).
- **164 tests passing** across the 5 changed test files.

## Test plan

- [ ] `npm test` passes locally.
- [ ] After merge auto-deploys: load `https://moderation.admin.divine.video`, click Refresh — confirm `Network` tab shows `?fresh=1` on `/admin/api/stats` and `/admin/api/ai-detection/stats`.
- [ ] Open a moderated video's detail view — confirm `client` and `content` still render.
- [ ] Untriaged feed: confirm shas that were `finished` then `deleted` do NOT appear (this requires production data; spot-check via D1 query if a moderator notices an obviously-deleted video back in the queue).
- [ ] `curl /api/v1/decisions?limit=1` (with auth) — confirm response shape is the original 8-column set, not the widened ADMIN_VIDEO_COLUMNS list.

## Rollback

Each fix is small. Reverting any individual concern is one or two-line.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/migrations/009-dashboard-speedup-indexes.sql
+++ b/migrations/009-dashboard-speedup-indexes.sql
@@ -24,4 +24,13 @@ CREATE INDEX IF NOT EXISTS idx_moderation_unreviewed
 -- Tracks the last time the backfill cron tried to populate this
 -- row's lookup metadata. Lets us skip rows we already tried for 7
 -- days (avoids hammering funnelcake on permanent 404s).
+--
+-- NOTE: SQLite does NOT support ADD COLUMN IF NOT EXISTS (D1 inherits
+-- this). Wrangler's migrations table prevents reruns under normal
+-- deploys, but if an operator ever has to manually re-apply this file
+-- (e.g. via `wrangler d1 execute --file=...`), the ALTER TABLE will
+-- error with "duplicate column name". That's the safe failure: nothing
+-- was already corrupted. Either drop the ALTER line for the manual
+-- replay, or split this concern into a separate migration file in a
+-- future change.
 ALTER TABLE moderation_results ADD COLUMN lookup_attempted_at TEXT;
--- a/src/admin/backfill-lookup-columns.mjs
+++ b/src/admin/backfill-lookup-columns.mjs
@@ -37,8 +37,14 @@ export async function runBackfill(env, options = {}) {
     return { skipped: 'disabled' };
   }
 
-  // Mutex: prevent the every-minute cron and a manual trigger from
-  // double-fetching the same shas.
+  // Best-effort mutex: prevents the every-minute cron and a manual
+  // trigger from double-fetching the same shas. Intentionally NOT
+  // strict — KV doesn't expose CAS, and read-then-write is racy under
+  // exactly-simultaneous calls. Two callers could both observe no
+  // lock and both proceed; the worst case is one duplicate batch of
+  // funnelcake calls (the UPDATE WHERE event_id IS NULL keeps the DB
+  // writes idempotent, so no data corruption). The 5-minute TTL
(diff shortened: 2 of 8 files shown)


DSanich (approved):
@rabble Great follow-up PR — overall this looks solid!

I ran tests locally and full suite passed.

PR #146, opened 2026-05-20
perf: fast-path public moderation status lookups

rabble:
## Summary
- Move public `/check-result/:sha` handling ahead of unrelated per-request schema initialization.
- Add a zero-DB `OPTIONS /check-result/:sha` fast path for public moderation status preflight.
- Validate and normalize public status SHA-256 path parameters before querying D1.
- Add regression tests proving public status lookup/preflight avoid schema setup and malformed hashes avoid DB access.

## Motivation
Moderation status checks are on the playback path and need to be fast for clients. The worker was doing unrelated D1 table/index initialization before serving public status lookups. While isolating that route, this also aligns public `/check-result/:sha` behavior with the legacy status endpoint so mixed-case hex hashes resolve correctly and malformed path values return `400` before any database work.

## Linked issue
Closes #176

## Manual validation
- [x] `npx vitest run src/index.test.mjs` — 149 passed
- [x] `npm run lint` — passed for 162 module files and 4 HTML files
- [x] `npx vitest run src/moderation/pipeline.test.mjs` — 23 passed
- [x] `npx wrangler deploy --dry-run` — succeeded
- [!] `npm test` — exited non-zero with Vitest worker `Timeout calling "onTaskUpdate"` after 350/363 tests and no assertion failures; the remaining `src/moderation/pipeline.test.mjs` file passes when run directly

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -1365,6 +1365,48 @@ async function handleLegacyStatus(sha256, env) {
   });
 }
 
+async function handlePublicCheckResult(url, env) {
+  const sha256 = url.pathname.split('/')[2];
+  const d1Result = await env.BLOSSOM_DB.prepare(`
+    SELECT sha256, action, provider, scores, categories, moderated_at, reviewed_by, reviewed_at, videoseal
+    FROM moderation_results
+    WHERE sha256 = ?
+  `).bind(sha256).first();
+
+  if (!d1Result) {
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz (approved):
Final pass complete. The public status fast path is scoped, covered, linked to #176, and CI is green.

PR #147, opened 2026-05-20
feat: add DELETE to validActions on /api/v1/moderate

mbradley:
## Summary

- Adds `DELETE` to the `validActions` array on the `/api/v1/moderate` endpoint
- `BLOSSOM_ACTION_MAP` already maps `DELETE` -> `DELETE` -- the only thing blocking DELETE requests was this validation array
- Prerequisite for relay-manager deletion alignment work (composable moderation actions, age review integration)

## Test plan

- [ ] `npx vitest run` passes (1,241 tests, all green)
- [ ] Deploy to production, verify `/api/v1/moderate` accepts `action: "DELETE"` for a test hash

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -3925,7 +3925,7 @@ async function runMigration() {
         }
 
         // Validate action
-        const validActions = ['SAFE', 'REVIEW', 'QUARANTINE', 'AGE_RESTRICTED', 'PERMANENT_BAN'];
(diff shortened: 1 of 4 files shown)


NotThatKindOfDrLiz (approved):
Final pass is clean. The follow-up fixes remove the action-list drift, make DELETE a first-class normalized moderation action, keep the admin endpoint intentionally non-DELETE, and add regression coverage for the relay-skip path. Local lint and the full test suite passed, and the required GitHub Actions checks are green on commit 3f374cc.

PR #148, opened 2026-05-21
feat: enable creator-delete pipeline in production

mbradley:
## Summary

- Flips `CREATOR_DELETE_PIPELINE_ENABLED` from `"false"` to `"true"` in `wrangler.toml`
- Already deployed to production — this PR syncs the toml with live state
- Prerequisites verified before deploy: D1 `creator_deletions` table with `physical_deleted_at` column (migrations 006+007), DELETE in `validActions` (PR #147)
- Blossom `ENABLE_PHYSICAL_DELETE` remains false — all deletes are soft-delete only

## Test plan

- [x] `delete-status` endpoint returns 400 (not 404), confirming the gate is open
- [x] Cron pipeline will begin processing kind-5 events on next tick

### DIFF
--- a/src/creator-delete/process.mjs
+++ b/src/creator-delete/process.mjs
@@ -102,7 +102,36 @@ export async function processKind5(kind5, { db, fetchTargetEvent, callBlossomDel
 
     const priorRetryCount = (claim.existing && claim.existing.retry_count) || 0;
 
-    const target = await fetchTargetEvent(target_event_id);
+    let target;
+    try {
+      target = await fetchTargetEvent(target_event_id);
+    } catch (fetchErr) {
+      const retryCountAfter = priorRetryCount + 1;
+      const status = retryCountAfter >= MAX_RETRY_COUNT
+        ? 'failed:permanent:max_retries_exceeded'
+        : 'failed:transient:target_fetch';
+      const lastError = fetchErr.message || 'Transient relay error fetching target event';
+      await updateToFailed(db, {
+        kind5_id: kind5.id,
+        target_event_id,
+        status,
+        last_error: lastError,
+        increment_retry: true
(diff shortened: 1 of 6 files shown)


dcadenas (changes requested):
This needs two fixes before it can merge.
The creator-delete path must prove the delete author owns the target event, and transient relay lookup failures must stay retryable instead of becoming permanent failures.

---
**Additional required changes**

- **[src/creator-delete/process.mjs:105](https://github.com/divinevideo/divine-moderation-service/blob/bd30bef9eb3303bd81d7d4a8a8f81b491511331a/src/creator-delete/process.mjs#L105)**
  Validate that `target.pubkey` matches `kind5.pubkey` before extracting the sha256 or calling Blossom DELETE.
  Right now a kind 5 event can reference another creator's video event, and this shared processor will still delete that target blob once the pipeline is enabled.
  Record a terminal author-mismatch failure and skip Blossom when the authors do not match, then cover both the mismatch and success paths in tests.
- **[src/creator-delete/process.mjs:105](https://github.com/divinevideo/divine-moderation-service/blob/bd30bef9eb3303bd81d7d4a8a8f81b491511331a/src/creator-delete/process.mjs#L105)**
  Do not turn every null `fetchTargetEvent` result into `failed:permanent:target_unresolved`.
  The production lookup also returns null for relay request failures or thrown fetches in `src/nostr/relay-client.mjs`, so a temporary outage, timeout, 429, 5xx, or indexing delay can make a valid delete stop retrying.
  Split confirmed not-found from transient lookup failure, and record transient lookup failures as retryable so cron can complete the delete later.

dcadenas (approved):
✅ creator-delete pipeline is ready

PR #150, opened 2026-06-04
feat(dm): add ACCOUNT_BANNED and ACCOUNT_RESTORED templates

mbradley:
Companion to divinevideo/divine-relay-manager#98; supports the moderation-DM coverage work in divinevideo/divine-relay-manager#96.

## What this does

Adds two account-state DM actions so relay-manager can distinguish account **ban**, **suspend**, and **restore**. Previously only `ACCOUNT_SUSPENDED` existed, so a ban rendered as a suspension.

## Changes

- **`src/index.mjs`** — added `ACCOUNT_BANNED` and `ACCOUNT_RESTORED` to the `/api/v1/notify` `validNotifyActions` allowlist.
- **`src/nostr/dm-sender.mjs`** — added matching no-arg `TEMPLATES` entries (mirroring `ACCOUNT_SUSPENDED`):
  - `ACCOUNT_BANNED` — permanent-action wording, with reply-to-appeal.
  - `ACCOUNT_RESTORED` — restoration wording, no appeal prompt, no content reference.
- `ACCOUNT_SUSPENDED` is unchanged.

## Tests

- `src/nostr/dm-sender.test.mjs` — asserts the message text for both new actions (and the negative cases: restore has no appeal/content reference).
- `src/index.test.mjs` — both actions pass the `/api/v1/notify` allowlist.

## Merge order

**Land this before relay-manager #98** so the new actions are accepted when relay-manager starts sending them. (If #98 lands first, those two DMs fail and are logged, non-critical, until this ships.)

CI green (tests + lint).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,8 +55,10 @@ jobs:
       - name: Run tests
         run: |
           set -o pipefail
+          set +e
           npm test 2>&1 | tee /tmp/vitest.log
           status=${PIPESTATUS[0]}
+          set -e
 
           if [ "$status" -eq 0 ]; then
             exit 0
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -3857,7 +3857,7 @@ async function runMigration() {
           });
         }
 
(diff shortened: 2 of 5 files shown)


NotThatKindOfDrLiz (approved):
Final pass complete. I pushed two small cleanups: a defensive selectTemplate guard for non-string actions, and a CI shell fix so the existing narrow Cloudflare storage-flake retry actually runs.\n\nValidated locally with lint, touched test suites, and full npm test; latest GitHub Lint/Test/Semantic PR/CLA checks are green. Ready to merge before the relay-manager companion PR.

PR #156, opened 2026-06-05
fix(moderation): route relay-admin actions through live /api/relay-rpc

mbradley:
## What
The moderation dashboard's "Ban User" button (and the content relay-deletion path) called the relay-admin `/api/moderate` endpoint, which is deprecated and no longer routed at the Cloudflare edge. It returns HTTP 522, so the request hung for the full edge timeout (~90s) and failed with only a faint status line (surfaced during moderator testing as an "hourglass that never finishes").

## Root cause
- `Ban User → /admin/api/uploader/:pubkey/enforcement → callRelayAdminAction → POST api-relay-prod.divine.video/api/moderate`.
- `/api/moderate` is deprecated (relay-manager's own code notes its `ban_pubkey` case "is not called by any frontend component"; the live admin UI uses `/api/relay-rpc`). Confirmed via `wrangler tail`: the relay-admin worker never receives `/api/moderate`, while `/api/relay-rpc` and sibling paths succeed.
- No timeout meant it hung the full edge timeout; the only feedback was a status line, so it read as a permanent hang.
- The same path backs content relay-deletion (`delete_event`), which was failing silently too.
- Verified the live path works by banning a throwaway test pubkey via the relay admin UI.

## Changes
- `src/index.mjs`: `callRelayAdminAction` now posts to `/api/relay-rpc`, mapping `ban_pubkey→banpubkey`, `allow_pubkey→unbanpubkey`, `delete_event→banevent`; adds a 15s AbortController timeout (fail fast).
- `src/admin/dashboard.html`: success/failure toast, button restored before the list refresh, and the refresh isolated so a refresh hiccup can't masquerade as a failed action.
- `src/uploader-enforcement.test.mjs`: assertions updated to the relay-rpc contract.

## Reviewer note (behavior)
Repointing `delete_event` to `/api/relay-rpc banevent` skips relay-manager's `handleModerate` side effects for that action (markHumanReviewed + Zendesk sync + a content DM). The moderation-service performs its own review bookkeeping and PERMANENT_BAN DMs, so this avoids a duplicate DM and redundant sync rather than losing coverage. Ban/unban have no side-effect change (handleModerate's `ban_pubkey` case just wraps the same `/api/relay-rpc` call).

## Out of scope (separate)
DM-sender WebSocket bug; the Cloudflare `/api/moderate` route cleanup; kind-1984/allowlist oddities.

## Test plan
- `node scripts/lint.mjs` passes; `vitest run` → 166 passing (incl. the `callRelayAdminAction` 302/CF-Access path).
- Manual after deploy: Ban User from the moderation dashboard completes in ~seconds with a toast.

### DIFF
--- a/src/admin/dashboard.html
+++ b/src/admin/dashboard.html
@@ -3441,19 +3441,29 @@
           throw new Error(data.error || `HTTP ${response.status}`);
         }
 
-        if (currentLookupVideo?.uploaded_by === pubkey) {
-          const reloadId = currentLookupVideo.lookupId || currentLookupVideo.sha256;
-          await lookupVideo(reloadId);
-        } else {
-          await loadVideos();
-        }
-
+        // Confirm success immediately (before the slower list refresh) so the
+        // moderator always gets a prompt, unmistakable signal it landed.
         button.innerHTML = originalText;
         button.disabled = false;
+        showToast(successMessage);
         setLookupStatus(successMessage);
+
+        // Refresh the affected view. A refresh hiccup must not look like the
+        // action itself failed, since it already succeeded above.
+        try {
+          if (currentLookupVideo?.uploaded_by === pubkey) {
+            const reloadId = currentLookupVideo.lookupId || currentLookupVideo.sha256;
+            await lookupVideo(reloadId);
+          } else {
+            await loadVideos();
+          }
+        } catch (refreshError) {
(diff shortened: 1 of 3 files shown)


dcadenas (approved):
🔌 ban actions hit the live relay-rpc now, no more 90s edge hang.

PR #164, opened 2026-06-05
feat(review): align untriaged queue + Needs-Triage count to the no-verdict definition (#158)

mbradley:
Implements the decision in #158: "needs triage" = videos the automation produced **no verdict** for (no `moderation_results` row), instead of the legacy KV `moderation:{sha}` key (written only by manual moderation / escalation / the transcript cron, **not** the `/api/v1/moderate` scan). The dashboard NEEDS TRIAGE count and the untriaged queue previously used different definitions and could drift; this unifies both on one anti-join.

See #158 for the decision and routing-model rationale.

**Note on current data:** `bunny_webhook_events` is empty in prod (0 rows; `moderation_results` ~324k). The untriaged queue is sourced from that table, so the queue and count read 0 in production today. The primary value of this change is **consistency** — the card and queue resolve from the same definition and can't drift. The reduced per-row fan-out described below is a structural property that only takes effect once the table is populated. Why the source table is empty is tracked in #168.

## Changes
- **`/admin/api/untriaged`** uses one anti-join (`latestUntriagedBunnyEvents`): latest bunny event per sha `LEFT JOIN moderation_results … WHERE m.sha256 IS NULL`. Filtering in the query (rather than post-filtering per row in app code) avoids per-row `MODERATION_KV.get('moderation:…')` reads and per-row relay enrichment, and makes `total` come from the same anti-join as the returned rows so the count and the list can't disagree.
- **Dashboard NEEDS TRIAGE count** uses the same anti-join (`countUntriagedBunnyEvents`) instead of the `totalInD1 − totalModerated` approximation, so the card and the queue can't drift.
- Adds `latestUntriagedBunnyEvents` / `countUntriagedBunnyEvents` (+ tests) and an untriaged timing log.
- **Triage-queue cards hydrate Nostr context client-side** (`dashboard.html`): the triage queue's render path (`loadTriageVideos` → `createTriageCard`) gets the same `#nostr-${sha}` container the moderated grid uses and hydrates each card via `loadNostrContext` after render (per-card, non-blocking), rather than relying on list-query enrichment. Also restores the pre-AI / Vine-era published badges.

## Behavior
The untriaged queue shows only no-verdict videos; AI-flagged items remain in the AI FLAGGED / Quick Review queue. Card and queue agree on one definition.

## Dead-code cleanup
`latestBunnyEventBySha` / `countLatestBunnyEvents` have no remaining app caller after this change, so they're removed. The rank-ALL-then-filter-the-winner regression (a finished-then-deleted sha must stay excluded) is ported into the untriaged test block, since the anti-join reuses the same `ranked` CTE.

Independent of #163 (`getAdminLookupVideo` parallelize) and #165. #163 originally shared the untriaged handler; this supersedes that hunk.

## Validation
- Full suite green (1298), lint clean. `bunny-events.test.mjs` covers the anti-join (only no-verdict shas; excluded once a row appears; deleted-status excluded; finished-then-deleted regression; LIMIT/OFFSET pagination). `dashboard-ui.test.mjs` pins the triage hydration hooks (`#nostr-${sha}` container, `divine-link` id, `loadNostrContext` call).
- Triage-card hydration is client-side `dashboard.html` JS (no unit harness); final visual confirmation at staging.

### DIFF
--- a/src/admin/bunny-events.mjs
+++ b/src/admin/bunny-events.mjs
@@ -8,74 +8,6 @@
 const DEFAULT_EXCLUDE_STATUS = ['error', 'deleted'];
 
 /**
- * Page through the latest event per sha, ordered most-recent-first.
- * Window-function CTE so each sha appears exactly once and the
- * non-aggregate columns (hls_url, mp4_url, etc.) come from the row
- * with the largest received_at — deterministic, unlike GROUP BY +
- * bare columns.
- */
-export async function latestBunnyEventBySha(env, options = {}) {
-  const {
-    limit = 50,
-    offset = 0,
-    excludeStatusNames = DEFAULT_EXCLUDE_STATUS,
-  } = options;
-  const placeholders = excludeStatusNames.map(() => '?').join(',');
-  // CRITICAL: rank ALL rows first, then filter by status_name on the
-  // winner. Filtering inside the CTE (before ROW_NUMBER) silently
-  // resurrects deleted shas — if a sha has finished@100 and deleted@200,
-  // dropping the deleted row from the CTE makes finished@100 the
-  // "latest". The old correlated subquery used MAX() over the full
-  // table and then required the latest row's status_name to be ok;
-  // this query mirrors that semantic.
-  const sql = `
-    WITH ranked AS (
-      SELECT
-        sha256, video_guid, hls_url, mp4_url, thumbnail_url, received_at, status_name,
-        ROW_NUMBER() OVER (PARTITION BY sha256 ORDER BY received_at DESC) AS rn
-      FROM bunny_webhook_events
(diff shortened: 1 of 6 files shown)


NotThatKindOfDrLiz (approved):
Approved after follow-up fix and final pass.\n\nI verified the no-verdict anti-join semantics, confirmed the lazy Nostr hydration now restores the backing triage card state and uploader context, and confirmed CI is green on head 8d7c9fc.\n\nValidation:\n- npm run lint\n- npm test -- src/admin/bunny-events.test.mjs src/admin/dashboard-ui.test.mjs src/index.test.mjs\n- npm test\n- git diff --check\n- GitHub checks: Lint, Test, semantic PR, CLA all green

PR #165, opened 2026-06-05
perf(review): cache + cached-host the per-card funnelcake lookup

mbradley:
Two latency wins on the per-card `/admin/api/video/{sha}` detail path, the work a moderator does on every card in the Quick Review queue. Both are do-no-harm: same data, just fewer/cheaper round-trips.

## #160 (host half) — read through the cached host
`fetchFunnelcakeLookupVideo` hit `relay.divine.video/api/videos/{id}`, the **uncached** backup path. Switched to the CDN-cached canonical host `api.divine.video` via a new `FUNNELCAKE_LOOKUP_URL` var (defaulted, env-overridable so staging can point elsewhere). Threaded `env` through the function's six callers.

Host check before changing anything (read-only, both hosts, same sample sha):
- Identical response bodies (18605 bytes, same shape).
- `api.divine.video` is genuinely edge-cached: `x-cache: MISS → HIT → HIT`, `x-cache-hits` incrementing.
- Safe TTL: `cache-control: public, max-age=15, stale-while-revalidate=45`, so review context is at most ~15-60s stale.

The NIP-98 admin **write** path (`FUNNELCAKE_ADMIN_URL`) stays on the relay and is untouched. (Note: the old hardcoded relay host was already prod, so staging already cross-read prod for these lookups; this doesn't change that, only makes the read cached.)

## #159 — cache the per-card lookup
`enrichAdminLookupVideo` fired an uncached funnelcake call on **every** detail-view click (`buildAdminVideoFromRow` hardcodes `client`/`content` to null, so `needsFunnelcake` is always true for flagged cards). Paging back and forth over the same handful of cards in a session re-fetched each one from scratch.

Now the successful result is cached in KV keyed by sha (`admin:lookup:funnelcake:{sha}`, 300s TTL). Repeat views skip the HTTP call entirely. Misses and failures are **never** cached, so a transient error can't pin "no context" for the window. The 300s TTL layered over the host's own 15-60s freshness keeps a review session stable without going meaningfully stale.

## Scope / no stacking
- This is the **cached-host half of #160** plus **#159**. The **parallelize half of #160** (the D1+KV `Promise.all` in `getAdminLookupVideo`) is in **#163**, independent — different lines, no overlap.
- Branched off `main`; touches only `fetchFunnelcakeLookupVideo` / `enrichAdminLookupVideo` / wrangler vars. No interaction with #164's untriaged/anti-join path.

## Validation
- Full suite green (1299, +1 for the new cache test), lint clean.
- New test: a second per-card request for the same sha hits the funnelcake REST endpoint exactly once across both views (Map-backed KV) and returns identical enriched context.
- Existing per-card tests updated to the new cached host.

Closes #159. Part of #160 (cached-host half here; parallelize half in #163) — close #160 once both land.

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -923,6 +923,9 @@ async function enrichAdminLookupVideo(video, env) {
     || (ctx.client == null && ctx.content == null);
 
   if (enriched.sha256 && needsFunnelcake) {
+    // Keyed on sha256 (the stable content hash) even though the fetch below may
+    // resolve via lookupId/eventId — the cached value is the record for that
+    // content hash regardless of which identifier resolved it.
     const cacheKey = `admin:lookup:funnelcake:${enriched.sha256}`;
     let funnelcakeVideo = null;
 
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -895,9 +895,11 @@ async function fetchLookupNostrContext(hash, env) {
 // back and forth over the same handful of cards in a review session — skips the
 // origin round-trip. Short TTL keeps title/author from going meaningfully stale
 // (the underlying api.divine.video response is itself only ~15-60s fresh).
-// Env-overridable so an operator can drop the staleness window during an
-// incident without a redeploy.
-const ADMIN_LOOKUP_CACHE_TTL_DEFAULT_SECONDS = 300;
+// Deliberately a fixed constant, not env-configurable: there's no operational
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz (approved):
Approved after final pass. The pushed follow-up closes the only concrete robustness gap I found: cache write failures now degrade like cache reads, with regression coverage. Local lint/full tests passed, and GitHub checks are green.

PR #166, opened 2026-06-05
perf(messages): one-pass conversation list + indexed thread lookup

mbradley:
Backend query optimization for the admin messages inbox — the first of the messages-perf PRs (do-no-harm, backend-only). Closes #154 and #155.

## #154 — conversation list: one grouped pass instead of per-row correlated COUNT
`getConversations` previously ran a per-returned-row correlated `(SELECT COUNT(*) … WHERE conversation_id = dl.conversation_id)` on top of a separate `WHERE id IN (SELECT MAX(id) … GROUP BY conversation_id)`. Rewritten to a single CTE that computes the latest row id **and** the message_count per conversation in one grouped scan, then joins back for the latest row's columns:

```sql
WITH latest AS (
  SELECT conversation_id, MAX(id) AS max_id, COUNT(*) AS message_count
  FROM dm_log GROUP BY conversation_id
)
SELECT … latest.message_count
FROM latest JOIN dm_log dl ON dl.id = latest.max_id
ORDER BY last_message_at DESC, dl.id DESC
LIMIT ? OFFSET ?
```

**No new index** (contra the issue's suggestion): `id` is `INTEGER PRIMARY KEY` (the rowid), so the existing `idx_dm_conversation` on `(conversation_id)` is physically `(conversation_id, id)` and already covers `MAX(id)`/`COUNT(*) GROUP BY conversation_id`. The final `ORDER BY` runs over the reduced one-row-per-conversation set (bounded by the moderation account's conversation count), so it doesn't warrant a `created_at` index either. Adding `(conversation_id, id)` would be a duplicate index — pure write/storage waste. Added an `id DESC` tiebreaker so pagination is deterministic when rows share a (1-second-resolution) `created_at`.

## #155 — thread lookup: deterministic id instead of unindexed OR scan
`getConversationByPubkey` filtered `WHERE sender_pubkey = ? OR recipient_pubkey = ?` — `sender_pubkey` is unindexed, so the OR forces a scan. The conversation id is deterministic, so the function now derives it via `computeConversationId(moderatorPubkey, pubkey)` and hits the indexed `conversation_id` column directly. Falls back to the OR query when no moderator pubkey is available (no signing key configured). The `/admin/api/messages/{pubkey}` caller passes the moderator pubkey (same `getModeratorPubkey(env)` it already uses for the list).

First of three messages-perf PRs, sliced by concern to avoid 3 PRs editing the same file:
- **This (A)** — `dm-store.mjs` query optimization (#154 + #155). Backend only.
- **B** — `messages.html` progressive list render + optimistic send (#152 + #151).
- **C** — thread pagination (#153), which spans both files; lands after A + B (depends on B's append-only render), rebased — not stacked.

## Validation
- Full suite green (1300), lint clean. The existing real-D1 `getConversations` suite validates the rewrite (per-conversation `message_count` correctness, correct `last_message` per conversation, limit/offset, empty). New real-D1 tests cover the `getConversationByPubkey` fast path (resolves the right thread via the deterministic id; null when the participant has no conversation).

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -3471,7 +3471,11 @@ async function runMigration() {
 
       const pubkey = url.pathname.split('/')[4];
       const { getConversationByPubkey } = await import('./nostr/dm-store.mjs');
-      const messages = await getConversationByPubkey(env.BLOSSOM_DB, pubkey);
+      const { getModeratorPubkey } = await import('./nostr/dm-reader.mjs');
+      // Pass moderatorPubkey so the lookup derives the deterministic
+      // conversation_id (indexed) instead of the unindexed sender/recipient OR scan.
+      const moderatorPubkey = env.NOSTR_PRIVATE_KEY ? getModeratorPubkey(env) : undefined;
+      const messages = await getConversationByPubkey(env.BLOSSOM_DB, pubkey, moderatorPubkey);
       if (!messages) {
         // Never-messaged recipient: return an empty thread (200) so the compose UI
         // can render for new conversations instead of showing a load error.
--- a/src/nostr/dm-store.mjs
+++ b/src/nostr/dm-store.mjs
@@ -121,10 +121,16 @@ export async function getConversationByPubkey(db, pubkey, moderatorPubkey) {
   }
 
   // Fallback (no moderator pubkey available, e.g. no signing key configured):
-  // find any conversation this pubkey participates in via the OR query.
+  // find the most recently updated conversation this pubkey participates in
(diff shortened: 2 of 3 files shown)


NotThatKindOfDrLiz (approved):
Approved. I did a final pass on the updated head (40c9d96): the query rewrite keeps the same conversation semantics, the moderator-aware fast path is scoped to the admin DM invariant, ordering tiebreakers are deterministic, and the fallback path is now deterministic too. Local lint, focused DM-store tests, and full test suite pass; GitHub checks are green.

PR #167, opened 2026-06-05
perf(messages): progressive conversation list + optimistic send

mbradley:
Frontend responsiveness for the admin messages inbox — second of the messages-perf PRs (`messages.html` only). Closes #152 and #151.

## #152 — render the conversation list before profiles resolve
`fetchConversations` `await`ed the bulk relay profile lookup (`fetchProfiles`, a kind-0 WS query with a 5s timeout) **before** the first `renderConversations()`, so the left panel sat on "Loading…" until the relay answered. Now it paints immediately — `displayName()` already falls back to the truncated pubkey, so rows show last message + timestamp right away — then fetches profiles in the background and re-renders to patch in names/avatars when they arrive.

## #151 — optimistic send
`sendMessage` disabled the composer, `await`ed the full publish round-trip (NIP-17 wrap + NIP-65 discovery + publish to every relay), then re-`loadThread`'d (a full DOM rebuild + scroll jump). Now it's optimistic:
- appends a **pending** outgoing bubble and clears/re-enables the composer immediately,
- publishes in the background,
- on success marks the bubble delivered (drops the pending state, stamps a timestamp) — no full-thread re-render,
- on failure **reverts**: removes the optimistic bubble and restores the typed text so the moderator can retry (your call: revert-and-restore over a persistent failed-retry affordance — simpler, less state).

Extracted `createMessageBubble` so the optimistic append (`appendOutgoingBubble`) and the server render (`renderThread`) share one bubble builder; `renderThread` is append-based and no longer full-rebuilds after each send. (Re-opening a thread still renders fresh from the server, so the optimistic bubble can't duplicate the logged row.)

Second of three messages-perf PRs. Independent of **A** (#166, `dm-store.mjs` query opt) — different file. **C** (thread pagination, #153) spans `dm-store.mjs` + `messages.html` and depends on this PR's append-based rendering, so it lands after A + B, rebased — not stacked.

## Validation
- Full suite green (1300), lint clean. `messages-ui.test.mjs` smoke tests pin the progressive-render reorder and the optimistic-send/revert wiring.
- Client-side `messages.html` JS (no unit harness for DOM behavior); final visual confirmation at staging (admin is CF-Access gated).

### DIFF
--- a/src/admin/messages-ui.test.mjs
+++ b/src/admin/messages-ui.test.mjs
@@ -26,3 +26,21 @@ describe('messages UI — new message compose hooks', () => {
     expect(messagesHTML).toContain('No messages yet');
   });
 });
+
+describe('messages UI — progressive render + optimistic send', () => {
+  it('renders the conversation list before profiles resolve (background patch)', () => {
+    // #152: paint immediately, then fetch profiles and re-render.
+    expect(messagesHTML).toContain('fetchProfiles(pubkeys).then(() => renderConversations())');
+  });
+
+  it('appends an optimistic pending bubble and reverts on failure', () => {
+    // #151: optimistic append + clear composer immediately, revert on error.
+    expect(messagesHTML).toContain('function createMessageBubble(');
+    expect(messagesHTML).toContain('function appendOutgoingBubble(');
+    expect(messagesHTML).toContain('appendOutgoingBubble(text, sha256, { pending: true })');
+    expect(messagesHTML).toContain('pending-tag');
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz (approved):
Approved after final pass.

I pushed `1f56abd` to preserve active conversation filters during background profile re-renders and to harden escaped display output in the shared message bubble builder. Local lint and full test suite passed, and GitHub CI is green on the updated head.

PR #171, opened 2026-06-09
fix(moderation): route relay-admin actions through a service binding + shared Secrets Store key (#170)

mbradley:
## Why

The per-card **Ban User** action calls the relay-admin worker's `/api/relay-rpc` over the public host `api-relay-prod.divine.video`, which returns **HTTP 522** (the CF edge can't reach the worker — reproduced live: toast `Relay admin error: HTTP 522`, log `[ENFORCE] Relay admin action failed: ... HTTP 522`). #156 did not resolve this; same edge-routing class as the old `/api/moderate` 522.

## What

Prefer a Cloudflare **service binding** (`RELAY_ADMIN` -> `divine-relay-admin-api-prod`) that invokes the worker by name — no edge, no DNS, no CF Access — authenticating with an `X-Admin-Key` header.

- The key is a **shared Cloudflare Secrets Store secret** (`MODERATION_TO_RELAY_ADMIN_KEY`, store `ef490704…`), bound here as `RELAY_ADMIN_API_KEY` and on the relay side as `MOD_RELAY_ADMIN_KEY` — one value, no duplication. Resolved in code via `.get()` (a plain string is also accepted for local dev).
- Falls back to the existing public-edge HTTPS + CF Access path when the binding is absent. The 15s fail-fast timeout and the failure behavior (caught -> HTTP 502 with the relay error, no enforcement persisted) are unchanged.

The relay-manager side that accepts this key: **divinevideo/divine-relay-manager#104** (must land + deploy first).

## Verified

- Reproduced the 522 live, then confirmed the fix targets the right layer (the 522 is the edge, not the relay worker — `handleRelayRpc` only ever returns 200/400).
- Independent code review: addressed (added binding-path timeout coverage); no outstanding issues.
- Lint clean; full suite **1311 tests pass**, including Secrets Store `.get()`, plain-string, binding-vs-fallback, and binding-timeout coverage.

## Deploy preconditions / order

1. `divine-relay-manager#104` deployed (so the relay worker accepts the key).
2. moderation-service's CI deploy token needs **Secrets Store read** to bind the secret (this worker has never used the store before — verify before merge).
3. Then merge this (CI auto-deploys to prod). Verify with a throwaway ban -> success.

## Follow-up once proven (scoped, not in this PR)

After the binding is proven in prod, a follow-up can collapse `callRelayAdminAction` to binding-only and delete the now-dead `getRelayAdminHeaders`. Note: the CF Access secrets themselves stay (also used by `src/nostr/publisher.mjs`), and relay-manager's `ADMIN_API_KEY` / CF Access stay (other callers + the admin UI). Keeping the fallback as a documented break-glass path is also a reasonable choice for a moderation-critical action.

Relates to #170.

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -766,6 +766,14 @@ function relayRpcForAction(payload) {
   }
 }
 
+// Resolve a secret that may be either a plain string (wrangler secret / [vars])
+// or a Cloudflare Secrets Store binding (async .get()). Mirrors the relay-admin
+// worker's resolveSecret so the shared key works regardless of how it is bound.
+async function resolveSecret(binding) {
+  if (!binding) return null;
+  return typeof binding === 'string' ? binding : await binding.get();
+}
+
 async function callRelayAdminAction(env, payload) {
   const rpcBody = relayRpcForAction(payload);
   const url = `${getRelayAdminUrl(env)}/api/relay-rpc`;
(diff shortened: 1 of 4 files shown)


NotThatKindOfDrLiz (approved):
Code review approved. Please hold merge until the manual deploy/Secrets Store confirmations in the PR thread are complete.

Issue #169, opened 2026-06-07
Apply kind-10050 DM routing + add Zendesk-reply→NIP-17 DM bridge (divine-mobile #570)


realmeylisdev:
Backend half of divine-mobile's NIP-17 moderation-DM work (divinevideo/divine-mobile#570). The mobile send path now resolves recipients' kind-10050 inbox relays (divinevideo/divine-mobile#4944); the server-side sender needs the same fix, plus the reply bridge.

### 1. Fix kind-10050 resolution in the DM sender
`dm-sender.mjs` (~lines 329–462) resolves recipient relays via **kind-10002 (NIP-65)**, which carries no DM-inbox semantics. Per NIP-17, gift wraps MUST be published to the recipient's **kind-10050** DM-relay list. As-is, outcome DMs to non-diVine users can be published to the wrong relays. Switch the resolution to kind-10050.

### 2. Zendesk-reply → NIP-17 DM bridge (TC-SOCIAL-026)
Net-new: when a T&S agent replies on a Zendesk ticket, deliver it to the reporting/reported user as a NIP-17 DM signed as the moderation account (`8fd5eb6d8f362163bc00a5ab6b4a3167dbf32d00ec4efdbcf43b3c9514433b7e`). Reuse the existing `sendModeratorReply` path; `divine-relay-manager` owns the Zendesk pre-auth + ticket↔event mapping and is the natural orchestration seam. (Per the #570 decision, the bridge lives in `divine-moderation-service`, not a new dedicated DM service.)

Context: full investigation + decisions in divinevideo/divine-mobile#570.

PR #173, opened 2026-06-16
feat(dm): publish moderation@ kind-10050 DM inbox relay list

mbradley:
## Summary

Publishes moderation@'s NIP-17 kind-10050 DM inbox relay list so strict NIP-17 clients can discover where to deliver gift-wrapped DMs for the moderation account.

Closes #174.

## Motivation

The moderation account already reads NIP-17 DMs from the configured home relay, but it did not advertise a kind-10050 DM inbox relay list. Per NIP-17, strict clients may refuse to DM a recipient without an advertised inbox relay list. This is the backend publish slice for the broader moderation DM work; it does not close the broader routing/bridge work in #169 or divinevideo/divine-mobile#4946.

## What Changed

- Added `publishDmInboxRelayList`, which signs a kind-10050 event with the existing `NOSTR_PRIVATE_KEY`.
- Advertises only `RELAY_POLLING_RELAY_URL` as the inbox relay, matching the relay the DM reader actually polls.
- Publishes the replaceable event to the home relay plus discovery relays for findability.
- Sends no Cloudflare Access credentials to relays, including third-party discovery relays.
- Added `DM_INBOX_PUBLISH_ENABLED`, defaulting to `"false"`, so this ships dormant.
- Added KV throttling so the static record publishes roughly daily only after the home relay accepts it.
- Cleaned up the publisher test seam so the tests exercise the real per-relay publish loop via injected `connect`.
- Added scheduled-handler coverage for flag gating, daily throttle, home-accept throttle advancement, home-reject retry behavior, and corrupt throttle timestamps.

## Sequencing / Operational Notes

Do not enable `DM_INBOX_PUBLISH_ENABLED="true"` until the production home relay accepts kind 10050. Before that, the home relay rejection is expected and the throttle intentionally does not advance.

No new required secrets or bindings are introduced. The implementation reuses `NOSTR_PRIVATE_KEY`, `RELAY_POLLING_RELAY_URL`, and `MODERATION_KV`.

## Validation

- `npm run lint`
- `npm test -- src/nostr/publisher.test.mjs src/index.test.mjs` (183 tests)
- `npm test` (74 files, 1320 tests)

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -7,7 +7,7 @@
 import { validateQueueMessage } from './schemas/queue-message.mjs';
 import { moderateVideo, classifyVideoOnly } from './moderation/pipeline.mjs';
 import { applyForceProvider, shouldQueueHiveRecheck } from './moderation/report-trigger.mjs';
-import { publishToFaro, publishToContentRelay, publishLabelEvent } from './nostr/publisher.mjs';
+import { publishToFaro, publishToContentRelay, publishLabelEvent, publishDmInboxRelayList } from './nostr/publisher.mjs';
 import { requireAuth, getAuthenticatedUser } from './admin/auth.mjs';
 import { verifyZeroTrustJWT } from './admin/zerotrust.mjs';
 import { getConfiguredBearerTokens, authenticateApiRequest, apiUnauthorizedResponse, authSourceFromVerification, verifyLegacyBearerAuth } from './auth-api.mjs';
@@ -5053,6 +5053,36 @@ async function runMigration() {
           console.log(`[CRON] DM inbox sync: ${syncResult.synced} new, ${syncResult.skipped} deduped, ${syncResult.errors} errors`);
         } catch (err) {
           console.error('[CRON] DM inbox sync failed:', err);
+        }
+      }
+
+      // Republish moderation@'s NIP-17 DM inbox relay list (kind 10050) ~daily so the
+      // account is reachable over standard NIP-17. Flag-gated so it can ship dormant
+      // until the relay accepts kind 10050 (divine-funnelcake#536); KV-throttled to
+      // avoid republishing this static record every cron tick.
+      if (env.DM_INBOX_PUBLISH_ENABLED === 'true') {
+        try {
+          const lastStr = env.MODERATION_KV
+            ? await env.MODERATION_KV.get('dm-inbox-relay-list:last-published')
+            : null;
+          const last = lastStr ? parseInt(lastStr, 10) : 0;
+          const dayMs = 24 * 60 * 60 * 1000;
+          if (Date.now() - last >= dayMs) {
+            const result = await publishDmInboxRelayList(env);
+            if (result.published) {
+              console.log(`[DM-INBOX] kind 10050 published to ${result.relays.length} relay(s); home relay ${result.homeRelayPublished ? 'accepted' : 'NOT yet (will retry next cron)'}${result.failed?.length ? `, ${result.failed.length} failed` : ''}`);
+              // Only throttle once the home relay (the sole inbox tag, and where we actually read
+              // DMs) accepted it. Discovery-relay success alone must not stop us retrying the relay
+              // that matters — e.g. before divine-funnelcake#536 the home relay rejects kind 10050.
+              if (result.homeRelayPublished && env.MODERATION_KV) {
+                await env.MODERATION_KV.put('dm-inbox-relay-list:last-published', String(Date.now()));
+              }
+            } else {
+              console.log(`[DM-INBOX] Publish skipped/failed: ${result.reason || 'all relays failed'}`);
+            }
+          }
+        } catch (error) {
+          console.error('[DM-INBOX] Republish error:', error.message);
         }
       }
 
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -5065,7 +5065,8 @@ async function runMigration() {
           const lastStr = env.MODERATION_KV
             ? await env.MODERATION_KV.get('dm-inbox-relay-list:last-published')
             : null;
-          const last = lastStr ? parseInt(lastStr, 10) : 0;
+          const parsedLast = lastStr ? parseInt(lastStr, 10) : 0;
+          const last = Number.isFinite(parsedLast) ? parsedLast : 0;
           const dayMs = 24 * 60 * 60 * 1000;
           if (Date.now() - last >= dayMs) {
             const result = await publishDmInboxRelayList(env);
@@ -5082,7 +5083,7 @@ async function runMigration() {
             }
           }
         } catch (error) {
-          console.error('[DM-INBOX] Republish error:', error.message);
+          console.error('[DM-INBOX] Republish error:', error?.message || String(error));
         }
       }
 
--- a/src/nostr/publisher.mjs
+++ b/src/nostr/publisher.mjs
@@ -336,3 +336,98 @@ function createLabelEvent(labelData, privateKeyHex) {
 
   return signedEvent;
 }
+
+/**
+ * Build a signed NIP-17 kind 10050 DM inbox relay list event.
+ *
+ * The `relay` tags advertise where this pubkey receives gift-wrapped (kind 1059)
+ * DMs. They must match where we actually read from (see dm-reader), or senders
+ * will deliver to relays we don't listen on.
+ *
+ * @param {string[]} inboxRelays - Relay URLs where moderation@ receives DMs.
+ * @param {string} privateKeyHex - Signing key (hex).
+ * @returns {Object} Signed kind-10050 event.
+ */
+function createDmInboxRelayListEvent(inboxRelays, privateKeyHex) {
+  const tags = inboxRelays.map((url) => ['relay', url]);
+
+  const unsignedEvent = {
+    kind: 10050,  // NIP-17 DM inbox relay list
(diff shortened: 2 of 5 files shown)


src/index.mjs
```diff
@@ -5056,6 +5056,36 @@ async function runMigration() {
         }
       }
 
+      // Republish moderation@'s NIP-17 DM inbox relay list (kind 10050) ~daily so the
+      // account is reachable over standard NIP-17. Flag-gated so it can ship dormant
+      // until the relay accepts kind 10050 (divine-funnelcake#536); KV-throttled to
+      // avoid republishing this static record every cron tick.
+      if (env.DM_INBOX_PUBLISH_ENABLED === 'true') {
```
realmeylisdev, line 5063: **Test coverage for the cron throttle wiring (non-blocking).** This block — the flag gate, the `Date.now() - last >= dayMs` comparison, and especially the `if (result.homeRelayPublished && env.MODERATION_KV)` guard that advances the throttle only when the home relay accepts — is the novel, load-bearing part of this change and has no test. `index.test.mjs` already drives `worker.scheduled(...)`, so a cron-level test is feasible: (a) flag off → no publish; (b) home accepts → KV advances; (c) home rejects + discovery succeeds → KV does **not** advance. The publisher function's failure case is covered, but the glue that actually reads/writes KV here is not.

src/index.mjs
```diff
@@ -5056,6 +5056,36 @@ async function runMigration() {
...
+        try {
+          const lastStr = env.MODERATION_KV
+            ? await env.MODERATION_KV.get('dm-inbox-relay-list:last-published')
```
realmeylisdev, line 5066: **Throttle edge cases (optional hardening).** Two theoretical edges: if `MODERATION_KV` were ever unbound, `last` becomes `0` so it publishes every tick and never persists; if the stored value were non-numeric, `parseInt` returns `NaN`, and `Date.now() - NaN >= dayMs` is false, so it would never publish again (stuck). Neither bites today (KV is always bound and the only writer stores `String(Date.now())`), but a `Number.isFinite(last)` guard would make it robust to corruption.

src/nostr/publisher.mjs
```diff
@@ -336,3 +336,104 @@ function createLabelEvent(labelData, privateKeyHex) {
...
+      try {
+        await relay.publish(event);
+        console.log(`[DM-INBOX] Published kind 10050 ${event.id} to ${url}`);
+        published.push(url);
+      } finally {
+        relay.close();
+      }
+    } catch (error) {
+      console.error(`[DM-INBOX] Failed to publish to ${url}:`, error.message);
+      failed.push({ relay: url, reason: error.message });
```
realmeylisdev: **Connection failures lose their reason (cosmetic).** `nostr-tools` rejects *connection-level* failures (unreachable / timeout) with a plain string rather than an `Error`, so `error.message` is `undefined` here — `failed[].reason` and the `console.error` will log `undefined` for an unreachable relay. Publish-level rejections (`OK:false`) do carry an `Error`, so the important "home relay rejected kind 10050" reason is preserved. Consider `error?.message || String(error)` for cleaner discovery-failure diagnostics.

src/nostr/publisher.mjs
```diff
@@ -336,3 +336,104 @@ function createLabelEvent(labelData, privateKeyHex) {
...
+  }
+
+  const published = [];
+  const failed = [];
+  for (const url of targets) {
+    try {
+      // relay.divine.video is a public Nostr relay that does not require CF Access for
+      // protocol traffic (matches the read path in dm-reader.mjs), and CF Access creds
+      // must never be sent to third-party discovery relays — so we send no headers here.
+      const relay = await connect(url);
```
realmeylisdev, line 409: **Pre-migration discovery cadence (minor).** Because the throttle only advances on home-relay success, if `DM_INBOX_PUBLISH_ENABLED` is flipped *before* the prod migration (funnelcake#536) lands, this loop republishes the identical replaceable event to the three public aggregators every 5 minutes indefinitely. The dormant default + documented sequencing mostly cover this, but a short throttle even on discovery-only success (or gating discovery publish on the post-dormancy window) would prevent hammering public relays if the flag is flipped early.

src/nostr/publisher.test.mjs
```diff
@@ -337,3 +337,79 @@ describe('publishLabelEvent (automated source)', () => {
...
+        tags: [['relay', 'wss://relay.divine.video']]
+      })
+    );
+
+    const event = mockRelay.publish.mock.calls[0][0];
+    expect(event.id).toEqual(expect.any(String));
+    expect(event.sig).toEqual(expect.any(String));
+    expect(event.pubkey).toEqual(expect.any(String));
+    expect(result.published).toBe(true);
+    expect(result.homeRelayPublished).toBe(true);
```
realmeylisdev, line 377: **Tautological happy-path assertion (non-blocking).** This test passes `mockRelay`, which hits the `if (mockRelay)` short-circuit that returns a hardcoded `homeRelayPublished: true` — so this assertion checks a constant, not the real loop. The home-relay-**success** path through the actual multi-relay loop is never asserted (only the home-reject case is, via the `connect`-injection test). An all-accept case using injected `connect` would close that gap and guard the `published.includes(homeRelay)` computation on the success side.

realmeylisdev (approved):
**Review: publish moderation@ kind-10050 DM inbox relay list**

I traced the correctness chain end-to-end and across the related repos (divine-funnelcake relay, the divine-mobile / divine-web consumers, and the discovery aggregators). The design is sound and the sequencing is correct. Approving; the inline notes are non-blocking.

**What I verified**
- **Identity is consistent end-to-end.** The kind-10050 signer (`publisher.mjs`), the outbound moderation DM sender (`dm-sender.mjs` `getModeratorKeys`), and the inbound DM reader (`dm-reader.mjs`) all derive from the same `env.NOSTR_PRIVATE_KEY`, so the advertised inbox belongs to the exact pubkey that actually reads DMs.
- **The inbox tag matches where we read.** Both the publisher and `dm-reader` resolve the home relay from `RELAY_POLLING_RELAY_URL` (default `wss://relay.divine.video`), so senders deliver gift wraps to a relay we poll.
- **Throttle safety holds.** `nostr-tools` `relay.publish()` rejects on `OK:false` / timeout, so before funnelcake#536 the home relay rejects kind-10050 and `homeRelayPublished` stays false, the KV throttle does not advance, and we keep retrying. I confirmed funnelcake#536 is still open/unmerged (kind-10050 is not yet in the relay allowlist in prod), so shipping dormant is correct and currently required.
- **NIP-17 shape is correct** (kind 10050, `['relay', url]` tags, empty content), and the "small inbox list, broadcast widely for discovery" model matches NIP-17 / NIP-65. The `'relay'` tag is exactly what divine-mobile parses (it accepts only `'relay'`) and divine-web accepts it too, so this unblocks mobile's already-shipped (currently inert) kind-10050 routing.
- **Discovery targets work:** purplepag.es and relay.damus.io accept and serve kind-10050 (checked against the live relays). I could not reach relay.nostr.band from my environment to confirm, so a one-off check there is worth doing; a single discovery relay dropping the event only affects findability, not the inbox itself.
- **Tests are green** (full suite passes; the determinism fix in 205abce is the right call).

**One doc nit:** the PR description says it "reuses the existing `NOSTR_NSEC`", but the code uses `NOSTR_PRIVATE_KEY`. `NOSTR_NSEC` / `getSecretKey` appear only in comments in this worker and refer to the separate relay-admin worker. The implementation is correct and consistent with the existing publisher/reader; only the description is slightly misleading.

src/index.mjs
NotThatKindOfDrLiz: Addressed in 38ce8d7: added scheduled-handler tests for flag off, home relay accepted with KV advancement, throttle not expired, and home rejected with no KV advancement.

src/index.mjs
NotThatKindOfDrLiz: Addressed in 38ce8d7: corrupt/non-numeric throttle timestamps now fall back to 0 with Number.isFinite, and the cron test covers recovery from a corrupt KV value.

src/nostr/publisher.mjs
NotThatKindOfDrLiz: Addressed in 38ce8d7: relay publish failures now use error?.message || String(error) for both logging and failed[].reason.

src/nostr/publisher.mjs
NotThatKindOfDrLiz: Leaving this as an operational sequencing guard rather than adding a second discovery throttle in this PR. The feature still ships dormant, and the PR body now explicitly says not to enable DM_INBOX_PUBLISH_ENABLED until the production home relay accepts kind 10050. That keeps the implementation focused and avoids adding extra state for a pre-enable failure mode.

src/nostr/publisher.test.mjs
NotThatKindOfDrLiz: Addressed in 38ce8d7: removed the mockRelay shortcut from publishDmInboxRelayList and rewrote the happy-path tests to use injected connect, so they now exercise the real per-relay loop and homeRelayPublished computation.

NotThatKindOfDrLiz (approved):
Final pass complete. The follow-up removes the redundant publisher seam, adds cron throttle coverage, hardens the timestamp/error handling, links the PR to closing issue #174, and all required checks are green.

PR #177, opened 2026-06-23
Age Review: correct the blossom action-to-state mapping comment

mbradley:
Relates to divinevideo/divine-relay-manager#110 and divinevideo/divine-relay-manager#117.

Documentation only. The comment claimed AGE_RESTRICTED maps to the withhold state; it
actually forwards AGE_RESTRICTED, which the media server maps to the age-gate state
(served to any signed-in viewer). Clarifies that the QUARANTINE -> RESTRICT path is the
reversible owner-only withhold. No code change.


NotThatKindOfDrLiz (approved):
Approved. I independently verified this is comment-only, matches the current divine-blossom webhook/access-control behavior, and local lint/tests pass.

PR #184, opened 2026-07-13
feat: authoritative community content-warning aggregation (#180)

### LINKED ISSUE #180: Authoritative aggregation + auto-apply + strikes for community content-warning tagging (divine-mobile #4771)
## Scope

Backend counterpart to the mobile slice of community content-warning tagging (divinevideo/divine-mobile#4771, PR divinevideo/divine-mobile#5720). The mobile app now lets viewers publish NIP-32 kind 1985 `content-warning` labels for a video and shows a **client-side, advisory** warning once 3 distinct Divine-NIP-05 authors agree. That client threshold is provisional. The authoritative half must live server-side. Good news: the primitives already exist across three repos; the new work is the aggregation trigger + escalation policy that wire them together.

## The pipeline (with existing primitives)

**1. Aggregate community votes → `divine-funnelcake`**
Community kind 1985 labels land in the relay; funnelcake's ClickHouse already understands this shape (`crates/clickhouse/src/client.rs` queries `kind = 1985` + the `content-warning` namespace; `event_filters.rs:20`; NIP-32 advertised in nip11). New work: count distinct, anti-abuse-weighted authors over kind 1985 by `#e`/`#a` target and detect the authoritative threshold crossing (stronger than the client's simple distinct-Divine-identity count: reputation, rate limits, account age, brigading detection).

**2. Auto-apply the authoritative label → this repo (`divine-moderation-service`)**
`src/nostr/publisher.mjs::publishLabelEvent()` already creates + signs (moderation account key) a **kind 1985 `content-warning`** label. When aggregation reports a crossing, call this publisher for the video. The mobile app already trusts and displays moderation-account labels via `moderation_label_service` (trusted-labeler path) — **no client change is needed to display an auto-applied tag.** Depends on divinevideo/divine-mobile#4946 (moderation account operational) and aligns with the trust decision in divinevideo/divine-mobile#4948.

**3. Strikes / blocking → `divine-relay-manager` admin API (+ `divine-blossom`)**
`divine-relay-manager` already ships `publishLabelAndBan` (`src/lib/adminApi.ts`, `LabelPublisher.tsx`) — publish a NIP-32 label and ban, today as a manual admin action. New work: a "repeated mislabel → strike → block" escalation policy on top of it; media-level restriction via divine-blossom `/admin/moderate` (BLOCK/RESTRICT).

## Constraints / notes

- The mobile client publishes **video-scoped** labels only (`e` + `a`, `content-warning` namespace) and deliberately **no `p` target** — a p-scoped 1985 label asserts an account-level content warning about the creator, not the video. Server-side account-level strike logic must be an explicit decision, not inferred from these video labels.
- This issue is the coordination home (the linchpin is this repo, which publishes the authoritative labels). Split out funnelcake / relay-manager sub-tasks as scoping firms up.

## Related

- Mobile slice: divinevideo/divine-mobile#4771, PR divinevideo/divine-mobile#5720
- Epic: divinevideo/divine-mobile#5177
- Moderation account operational: divinevideo/divine-mobile#4946
- Trust model: divinevideo/divine-mobile#4948

mbradley:
## Summary

Authoritative backend for community content-warning tagging. A `*/5` cron sweep aggregates community NIP-32 kind 1985 `content-warning` votes from the relay, and once a **KV-configurable threshold (default 3) of distinct Divine-NIP-05-identity authors** agree on a label for a video, the moderation account publishes **one authoritative kind 1985 label** (`e` + `x` targets). Clients already render moderation-account labels through their existing trusted-labeler path, so no client changes are needed.

Creators accumulate a **strike** when the community had to label a video they did not self-label; at a configurable strike level the moderation account sends a **warning DM**. **Bans are never automated**: `GET /admin/api/community-strikes` ranks repeat offenders for human review using existing ban tooling.

Closes #180. Related: divinevideo/divine-mobile#4771 (client half, PR divinevideo/divine-mobile#5720), divinevideo/divine-mobile#5177 (epic).

## Behavior changes

- **New relay publishing:** the moderation account emits kind 1985 `content-warning` labels with `source: community` metadata when community consensus crosses the threshold.
- **New moderation outcome:** per-creator strike accounting plus automated warning DMs via the existing rate-limited moderator DM path. No automated bans, quarantines, or deletions.
- **Everything is inert until activated:** the whole pipeline sits behind KV `community_labels_enabled` (default off). Deploying this PR changes nothing at runtime until the switch is enabled.

## Design

Spec and plan are committed under `docs/superpowers/`. `src/community-labels/decision.mjs` is pure (no I/O), which is the seam a future Osprey rule replaces while the actuator (publisher, DM, D1 audit tables, admin surface) survives that rebuild. The KV kill switch doubles as the Osprey cutover lever.

Pipeline: poll kind 1985 since a watermark cursor -> group by target video with each video's earliest new vote -> process oldest-first, batch-capped (default 50) -> per video: fetch the video event (transient errors hold the cursor; genuine missing event skips) -> full per-video tally with explicit relay cap/fail-closed behavior -> Divine-identity gate (name-server `by-pubkey`, KV-cached 24h with KV expiration; transient failures throw and hold the cursor) -> publish-once via D1 PK dedup -> strike/warning accounting.

The cursor advances to just before the earliest vote of any deferred or failed video, so no vote is ever left behind it unprocessed. A full possibly-truncated poll page holds and freezes the cursor so newest-first relay truncation cannot drop the oldest votes.

## Settings (KV)

| Key | Default | Meaning |
|---|---|---|
| `community_labels_enabled` | off | Kill switch; only exact `'true'` runs the sweep |
| `community_label_threshold` | 3 | Distinct Divine-identity reporters to auto-apply |
| `strike_warning_count` | 3 | Strikes before the warning DM |
| `community_sweep_batch_limit` | 50 | Max videos evaluated per tick |

## Security posture and v1 limitations

v1's authoritative trigger is intentionally the **same** simple distinct-Divine-identity count the client uses: nothing stronger. Because Divine NIP-05 registration is open, a small number of minted identities (default 3) can force an authoritative label and a creator strike, with no velocity, account-age, or brigading guard. This is acceptable **only** because the pipeline is kill-switched, strikes never auto-ban (human review), and hardening is deferred to the Osprey rebuild. **The T&S sign-off before enabling in prod should treat this Sybil vector as a known, accepted v1 limitation.**

High-volume residuals (batch-boundary ties, permanently-failing-oldest starvation, and sustained full relay pages) are documented as accepted, observable limitations with page-backward and fairness/dead-letter fast-follows.

## Validation

- `npm run lint`: clean.
- `npm test`: 1418 passing.
- Migration `010-community-labels.sql` is additive only.
- Manual rollout: apply `migrations/010-community-labels.sql` to the target D1 database -> deploy dark -> confirm no `[COMMUNITY-LABELS]` activity with the switch off -> enable in staging KV -> publish 3 test votes from Divine-identity accounts -> verify one authoritative label appears and the mobile client blurs -> verify strike row plus warning DM with `strike_warning_count=1` -> enable in prod KV when signed off.


src/community-labels/sweep.mjs
```diff
@@ -0,0 +1,253 @@
...
+  const targets = [...touched.values()].sort((a, b) => a.earliestVoteAt - b.earliestVoteAt);
+  const batch = targets.slice(0, batchLimit);
+  // Earliest vote of every video the watermark must not pass: batch-cap
+  // deferrals up front, per-video failures appended below.
+  const unprocessedAt = targets.slice(batchLimit).map((target) => target.earliestVoteAt);
+
+  for (const target of batch) {
+    let videoClean = true;
+    try {
+      const videoEvent = await fetchVideoEvent(target.eventId);
```
dcadenas, line 151: At `src/community-labels/sweep.mjs:141`, validate the fetched target before tallying votes. Skip events whose kind is not 34235 or 34236, and derive the addressable id from the fetched video's own `kind:pubkey:d` tag instead of `target.addressableId`. Otherwise a public vote can pair a victim's `e` with another video's `a`, borrow that video's consensus, and publish a label or strike against the victim.

src/community-labels/sweep.mjs
```diff
@@ -0,0 +1,253 @@
...
+      }
+
+      const crossings = decideCrossings(votesByLabel, divineByAuthor, threshold);
+      const selfLabels = creatorSelfLabels(videoEvent);
+      const sha256 = sha256Of(videoEvent);
+
+      for (const crossing of crossings) {
+        if (await hasDecision(db, target.eventId, crossing.label)) continue;
+
+        const publishResult = await publishLabel({
```
dcadenas: Make the side effects at `src/community-labels/sweep.mjs:170` crash-safe and independently retryable. Persist or atomically claim the decision before publishing, keep a published event id for deduplication, and record strikes through their own idempotent check instead of skipping them when `hasDecision` is true. Apply the same marker-before-send ordering to warning DMs, so an isolate or D1 failure cannot create duplicate labels or DMs or permanently lose a strike.

src/community-labels/identity.mjs
```diff
@@ -0,0 +1,58 @@
...
+      }
+    } catch {
+      // Corrupt cache entry: fall through to a fresh lookup.
+    }
+  }
+
+  let found;
+  try {
+    const response = await fetchImpl(`${BY_PUBKEY_BASE}/${normalized}`);
+    if (response.status !== 200) return false;
```
dcadenas: At `src/community-labels/identity.mjs:37`, distinguish transient lookup failures from a confirmed non-Divine result so the sweep holds the watermark and retries. Make `queryRelay` reject a `collectAll` subscription that closes before EOSE, and treat 401/403 responses and malformed successful event bodies as transient under `throwOnTransient`. These failures must reach the existing hold/retry path instead of advancing past votes that were not fully evaluated.

src/community-labels/d1.mjs
```diff
@@ -0,0 +1,83 @@
...
+     VALUES (?, ?, ?)
+     ON CONFLICT(creator_pubkey, strike_count) DO NOTHING`
+  ).bind(creatorPubkey, strikeCountValue, now).run();
+}
+
+/**
+ * Creators ranked by strike count (desc) for the admin review feed.
+ * Returns [{ creator_pubkey, strikes, last_at }].
+ */
+export async function listStrikeSummary(db, { limit = 100 } = {}) {
```
dcadenas: Extend `src/community-labels/d1.mjs:74` and the admin response to include the rows behind each creator's count, at least `video_event_id`, `label`, and `created_at`, or add a drill-down keyed by creator. The database already stores this data, but returning only `COUNT(*)` and `MAX(created_at)` leaves moderators unable to verify which videos caused a ban-review score.

src/community-labels/decision.mjs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ * Content-warning labels the creator applied on the video event itself,
+ * normalized: NIP-32 self-labels plus bare content-warning tags.
+ */
+export function creatorSelfLabels(videoEvent) {
+  const labels = new Set();
+  for (const tag of videoEvent?.tags ?? []) {
+    if (!Array.isArray(tag) || tag.length < 2) continue;
```
dcadenas: Handle a bare `['content-warning']` tag at `src/community-labels/decision.mjs:85` as the generic `content-warning` self-label. Keep specific-category matching unchanged, and add coverage showing the generic crossing produces no strike for that tag while unmatched specific labels still do.

dcadenas (dismissed):
Before merge, make the sweep validate that each target is a video, use the fetched video's own address, and make publishing, strikes, and warning DMs retry-safe. Preserve the watermark across transient dependency failures, expose the stored per-strike evidence in the admin feed, and credit bare generic self-warnings so moderation results are not forged, lost, or misapplied.

src/community-labels/sweep.mjs
```diff
@@ -0,0 +1,253 @@
...
+    // Touched-video detection keys on the `e` (event id) target only,
+    // though the per-video tally below fetches both `e` and `a`. The whole
+    // downstream is event-id-keyed (fetchVideoEvent by id, decision PK on
+    // video_event_id), so an `a`-only vote has no concrete event to label;
+    // the mobile client always co-tags `e`+`a`, so real votes carry an `e`.
+    if (target.eventId === null) continue;
+    const createdAt = Number.isInteger(vote?.created_at) ? vote.created_at : now;
+    const existing = touched.get(target.eventId)
+      ?? { eventId: target.eventId, addressableId: null, earliestVoteAt: createdAt };
+    existing.addressableId ??= target.addressableId;
```
NotThatKindOfDrLiz: This carries forward the vote event's `a` tag, which is attacker-controlled, and later uses it in the full tally for the `e` target. A single forged vote can pair a victim video `e` with some unrelated addressable video `a` that already has threshold community votes, causing us to publish an authoritative label and record a strike against the victim creator. Please derive the addressable id from the fetched video event itself (`kind:pubkey:d`) after validating the target, not from the vote.

src/community-labels/sweep.mjs
NotThatKindOfDrLiz, line 151: `fetchVideoEvent` returns any Nostr event by id; this path should skip anything that is not a Divine video kind (`34235` or `34236`) before tallying votes or assigning strikes. Otherwise a label vote targeting a profile/note/etc. can turn that event's author into the punished creator.

src/community-labels/identity.mjs
NotThatKindOfDrLiz: A transient name-server failure is currently indistinguishable from a confirmed non-Divine result, so the sweep can under-count real votes, decide there is no crossing, and advance the watermark past those votes permanently. The sweep needs a mode where transient identity failures throw and hit the existing hold/retry path; only definitive not-found results should return `false`.

src/nostr/relay-client.mjs
```diff
@@ -453,6 +453,31 @@ export async function fetchKind5EventsSince(sinceSeconds, relayUrl = 'wss://rela
...
 }
 
+/**
+ * Fetch kind 1985 (NIP-32 label) events created since a cursor. Callers
+ * filter by namespace in code; relays are not assumed to index #L.
+ */
+export async function fetchLabelEventsSince(sinceSeconds, relayUrl = 'wss://relay.divine.video', env = {}, { limit } = {}) {
+  const filter = { kinds: [1985], since: sinceSeconds };
+  if (Number.isInteger(limit) && limit > 0) filter.limit = limit;
+  return queryRelay(relayUrl, filter, env, { collectAll: true });
```
NotThatKindOfDrLiz: This label fetch depends on `queryRelay(..., { collectAll: true })`, whose `close` handler resolves with whatever events arrived even if no EOSE was seen. A mid-stream socket close therefore looks like a complete below-threshold result and lets the watermark advance. Please make pre-EOSE collect-all closes reject under the sweep/label-fetch transient mode so these votes are retried instead of silently dropped.

src/community-labels/sweep.mjs
```diff
@@ -0,0 +1,253 @@
...
+      for (const author of authors) {
+        divineByAuthor.set(author, await isDivine(author));
+      }
+
+      const crossings = decideCrossings(votesByLabel, divineByAuthor, threshold);
+      const selfLabels = creatorSelfLabels(videoEvent);
+      const sha256 = sha256Of(videoEvent);
+
+      for (const crossing of crossings) {
+        if (await hasDecision(db, target.eventId, crossing.label)) continue;
```
NotThatKindOfDrLiz: This short-circuit makes the strike side effect non-retryable. If publish and `recordDecision` succeed but `recordStrike` throws, the next sweep sees the decision and skips the crossing forever, permanently losing the strike. Decisions, strikes, and warnings need independent idempotent recovery instead of using `hasDecision` as the gate for all downstream work.

src/community-labels/decision.mjs
NotThatKindOfDrLiz: A bare `['content-warning']` tag is a valid generic self-warning signal but is dropped by this `tag.length < 2` guard. At minimum, credit it as the generic `content-warning` self-label, and add coverage that a generic crossing does not produce an unearned strike while unmatched specific labels still do.

src/community-labels/sweep.mjs
```diff
@@ -0,0 +1,253 @@
...
+            videoEventId: target.eventId,
+            label: crossing.label,
+            now,
+          });
+          summary.strikes += 1;
+        }
+      }
+
+      const strikes = await strikeCount(db, videoEvent.pubkey);
+      if (strikes >= warningCount && !(await warningSent(db, videoEvent.pubkey, strikes))) {
```
NotThatKindOfDrLiz: This sends once per exact strike count, not once per escalation level. With the default threshold of 3, a creator can receive warnings at 3, 4, 5, ... strikes. Gate this on explicit escalation levels, such as multiples of `warningCount`, or store a normalized warning level instead of the raw current count.

src/community-labels/sweep.mjs
```diff
@@ -0,0 +1,253 @@
...
+      videoClean = false;
+    }
+    if (!videoClean) unprocessedAt.push(target.earliestVoteAt);
+  }
+
+  const watermark = unprocessedAt.length === 0
+    ? watermarkCeiling
+    : Math.min(Math.min(...unprocessedAt) - 1, watermarkCeiling);
+
+  if (watermark > cursor) {
```
NotThatKindOfDrLiz: The full-page cold-start freeze is handled below, but the failure path still has the same sliding-window hole. On a first run, if a video fails and the computed watermark is at or below the unpersisted default cursor, this branch writes nothing; the next tick's default `now - 24h` can slide past the failed vote. Persist the held cursor/watermark for failure holds too, not only for full pages.

src/community-labels/d1.mjs
```diff
@@ -0,0 +1,83 @@
...
+  const { results } = await db.prepare(
+    `SELECT creator_pubkey, COUNT(*) AS strikes, MAX(created_at) AS last_at
```
NotThatKindOfDrLiz, line 205: The admin feed returns only counts and a last timestamp, but the design depends on human review before bans. Moderators need the evidence behind each score, at least `video_event_id`, `label`, and `created_at`, either inline in this response or via a drill-down endpoint keyed by creator.

src/index.mjs
```diff
@@ -5239,6 +5260,47 @@ async function runMigration() {
...
+                score: 1,
+                source: 'community',
+                voteCount,
+                nostrEventId: videoEventId,
+              }, env);
+              return { published: result.published === true, eventId: result.eventId };
+            },
+            sendWarningDm: async ({ creatorPubkey, strikeCount, videoSha256 }) => {
+              const message = `Heads up from Divine moderation: your account has ${strikeCount} content-warning strikes. Community consensus applied warning labels to videos you posted without them. Please add content warnings when sharing sensitive content. Repeated omissions are reviewed by our moderators and can lead to account restrictions.`;
+              return sendModeratorReplyDm(creatorPubkey, message, videoSha256, env, ctx);
```
NotThatKindOfDrLiz: This automated warning goes through `sendModeratorReply`, whose audit path records `messageType: 'moderator_reply'` and whose docstring says it is for manual appeal responses. Please add/use a dedicated community-warning DM path or template so automated strike warnings are distinguishable from human moderator replies in the DM/audit trail.

NotThatKindOfDrLiz (changes requested):
Requesting changes. This is a moderation actuator that publishes under the moderation identity and records punitive strike state, so the trust-boundary and retry semantics need to be tightened before merge. The kill switch is useful for rollout, but it should not be the thing preventing these cases from reaching production behavior.

NotThatKindOfDrLiz (approved):
Approved after takeover pass. The remaining review blockers have been addressed in db7ff2d, PR body/issue linkage are corrected, and local plus GitHub verification are green.

src/community-labels/sweep.mjs
```diff
@@ -0,0 +1,287 @@
...
+      const selfLabels = creatorSelfLabels(videoEvent);
+      const sha256 = sha256Of(videoEvent);
+
+      for (const crossing of crossings) {
+        // Publish once: skip when an authoritative label already exists. But
+        // the strike (and warning below) are ensured independently every sweep
+        // — a strike that failed to land after its decision was recorded must
+        // be recoverable, not lost forever behind this publish-once dedup.
+        if (!(await hasDecision(db, target.eventId, crossing.label))) {
+          const publishResult = await publishLabel({
```
dcadenas: Persist the label event and its stable ID before calling `publishLabel`, then reuse that event on retry and mark it sent afterward. The current `publishLabel`-then-`recordDecision` order at `src/community-labels/sweep.mjs:196-215` can publish successfully, lose the D1 marker, and publish a duplicate when the held cursor retries. Apply the same durable pending or claim-before-send pattern to `sendWarningDm` and `recordWarning` at lines 244-251.

src/community-labels/d1.mjs
```diff
@@ -0,0 +1,136 @@
...
+ * per creator. Returns a Map<creator_pubkey, [{ video_event_id, label,
+ * created_at }]>. One query for the whole page rather than N per-creator ones.
+ */
+const D1_MAX_BOUND_PARAMETERS = 100;
+
+async function listStrikeDetails(db, { creatorPubkeys, perCreatorLimit }) {
+  const byCreator = new Map();
+  if (!Array.isArray(creatorPubkeys) || creatorPubkeys.length === 0) return byCreator;
+
+  for (let i = 0; i < creatorPubkeys.length; i += D1_MAX_BOUND_PARAMETERS) {
```
dcadenas, line 170: Add an authenticated, creator-keyed, paginated drill-down for every strike behind the displayed total. `listStrikeDetails` caps the preview at 20 rows at `src/community-labels/d1.mjs:84-105,117-134`, while the admin endpoint has no route for older rows, so a creator with 25 strikes leaves five count-contributing videos, labels, and timestamps unauditable. Query only the requested page in SQL and verify that paging can retrieve every underlying strike.

dcadenas (dismissed):
The relay and warning side effects are still vulnerable to duplicate delivery when D1 fails after the network action. Before merge, preserve retryability for transient or malformed video fetches and expose every strike behind the displayed total through an authenticated, paginated drill-down.

---
**Additional required changes**

- **Keep transient video-fetch responses retryable** ([src/nostr/relay-client.mjs:523](https://github.com/divinevideo/divine-moderation-service/blob/db7ff2d2b6b5004e922df8f5878815660388e386/src/nostr/relay-client.mjs#L523))
  Treat only a real 404 as definitive absence, and validate a successful response body before marking it definitive. At `src/nostr/relay-client.mjs:523-537`, 401/403 responses, invalid JSON, and malformed 2xx bodies resolve to `null` under `throwOnTransient`, so the sweep can advance its watermark as if the video were missing. Add coverage for each retryable case and verify a real 404 still resolves to `null`.

src/nostr/relay-client.mjs
```diff
@@ -474,13 +521,22 @@ export async function fetchNostrEventById(eventId, relays = ['wss://relay.divine
...
         continue;
       }
-      anyDefinitiveResponse = true;
+      // A 2xx counts as definitive only once we have a real event body.
+      // Invalid JSON throws into the catch below (transient); a malformed
+      // 2xx missing id/pubkey falls through to `continue` (transient).
+      // Neither is treated as "event absent".
       const event = await response.json();
-      if (event?.id && event?.pubkey) return event;
+      if (event?.id && event?.pubkey) {
```
dcadenas: Validate the relay response as the requested signed event before treating it as definitive. Require `event.id === eventId`, canonical ID/signature validation, and valid field shapes for everything the sweep consumes. Under `throwOnTransient`, route mismatched or malformed 2xx bodies through the transient-error path so the sweep cannot publish a label for one event while striking or messaging another.

src/community-labels/sweep.mjs
```diff
@@ -0,0 +1,313 @@
...
+      const crossings = decideCrossings(votesByLabel, divineByAuthor, threshold);
+      const selfLabels = creatorSelfLabels(videoEvent);
+      const sha256 = sha256Of(videoEvent);
+
+      for (const crossing of crossings) {
+        // Publish once: skip when a CONFIRMED authoritative label already
+        // exists. But the strike (and warning below) are ensured independently
+        // every sweep — a strike that failed to land after its decision was
+        // recorded must be recoverable, not lost forever behind this dedup.
+        if (!(await hasDecision(db, target.eventId, crossing.label))) {
```
dcadenas, line 198: Persist and reuse the exact prepared actuator event across retries instead of reconstructing it from partial claim data. The label claim currently freezes only `created_at` and `vote_count`, so a publish/confirmation split can produce a new label ID after code or key changes. Also retain the warning claim for ambiguous post-send outcomes: `sendWarningDm()` can return false after `EVENT` was accepted, but `src/community-labels/sweep.mjs:274` releases the claim and allows a fresh gift wrap on the next tick.

dcadenas (dismissed):
The moderation outcomes are not yet safe to merge because recovery can still split event identity or repeat a warning after an ambiguous send. Address both inline comments before merging.

src/community-labels/sweep.mjs
```diff
@@ -0,0 +1,325 @@
...
+        if (claimed) {
+          const dmResult = await sendWarningDm({
+            creatorPubkey: videoEvent.pubkey,
+            strikeCount: strikes,
+            videoSha256: sha256,
+          });
+          if (dmResult?.sent) {
+            await confirmWarning(db, { creatorPubkey: videoEvent.pubkey, warningLevel });
+            summary.warned += 1;
+          } else {
```
dcadenas, line 294: Release the pending warning claim for definitive pre-send failures, while retaining it only when delivery became ambiguous. `src/nostr/dm-sender.mjs:872-875` returns `{sent:false}` for rate limiting before the gift-wrap is created or published, but this branch retains every false result, so a later sweep can never retry that warning. Add regression coverage for both a proven no-send retry and an ambiguous post-publication failure that must not resend.

dcadenas (dismissed):
The warning-delivery path still permanently consumes a warning claim when the DM sender rejects before creating or publishing a message. Before merge, distinguish definitive pre-send failures from ambiguous post-send outcomes so rate-limited warnings can be retried without allowing duplicate warnings after uncertain relay delivery.

src/nostr/dm-sender.mjs
```diff
@@ -820,65 +835,82 @@ export async function notifyReporters(sha256, action, env, logPrefix = '[DM]') {
...
       message.trim()
     );
 
     const relayUrls = await discoverUserRelays(recipientPubkey, env);
+    // From here a relay may accept the event even if we never see the OK, so
+    // any failure at or after this point is ambiguous (definitive: false).
+    publishAttempted = true;
     const { success } = await publishToRelays(wrappedEvent, relayUrls, env);
 
     if (success === 0) {
```
dcadenas, line 918: Preserve the distinction between explicit relay rejection and an unknown delivery outcome here. `publishToRelays` currently reduces both cases to `success === 0`, so `sendModeratorMessage` returns `definitive: false` at this line and the sweep keeps the pending claim even when every relay returned `OK=false`; that warning level can never be retried. Return a definitive failure for all-explicit-rejection results, while keeping timeout, socket-error, and close-without-`OK` results ambiguous, and update the existing all-`OK=false` test accordingly.

dcadenas (changes requested):
Complete the relay outcome handling before merge. The sender must distinguish explicit `OK=false` rejections from missing acknowledgements so a warning claim is released after every relay rejects the event, while timeout, socket-error, and close-without-ack paths remain retry-safe.

dcadenas (approved):
✅ community warnings aggregate cleanly, ship it

PR #185, opened 2026-07-13
chore(dm): enable moderation@ kind-10050 DM inbox publish

mbradley:
flips DM_INBOX_PUBLISH_ENABLED to true so moderation@ advertises its NIP-17 DM inbox (kind 10050, ~daily republish, KV-throttled).

the flag's own stated precondition is met: divine-funnelcake#536 is live in prod, verified 2026-07-13 (the prod relay stores kind-10050 events from ordinary users). the dm-reader already polls the inbox unconditionally every cron tick, so this only starts advertising an inbox we already read. the event is replaceable, so this is reversible by publishing a replacement.

also corrects the config comment: cloudflare rejects a secret shadowing a deployed [vars] binding name, so this flag flips via var change + deploy, not `wrangler secret put`.

part of divinevideo/divine-mobile#4946.

### DIFF
--- a/wrangler.toml
+++ b/wrangler.toml
@@ -170,11 +170,14 @@ BACKFILL_ROWS_PER_TICK = "200"
 # Creator-delete pipeline feature flag (default off; flip to "true" via `wrangler secret put` or [vars] once ready)
 CREATOR_DELETE_PIPELINE_ENABLED = "true"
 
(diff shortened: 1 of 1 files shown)


dcadenas (approved):
✅ moderation DMs now reach the inbox

PR #188, opened 2026-07-24
fix(creator-delete): avoid missing delayed kind-5 events

dcadenas:
## Summary

Delayed kind-5 deletion events could be missed forever when relay indexing finished after the cron checkpoint advanced.
This change overlaps persisted checkpoint polls by five minutes, treats truncated relay polls as failed polls, and keeps replay safe.
Transient failures stay on the existing exponential-backoff retry path per target, so one transient target does not hide sibling targets from the same kind 5.

## Motivation

The cron queried from the exact last successful poll time.
An event created earlier but indexed later fell behind every later `since` filter.
A bounded overlap covers the observed delay plus cron alignment while keeping replay work bounded.

## Related Issue

- Closes #187

## Validation

The automated cron regressions, full suite, and lint checks pass on the takeover commit.

- [x] `npm run lint` - 179 modules and 4 HTML files
- [x] `npm test` - 81 files and 1,451 tests
- [x] relevant Wrangler or queue-flow validation if applicable - creator-delete cron tests cover delayed visibility, replay idempotency, per-target retry backoff, sibling target processing, and strict relay truncation handling
- [ ] I could not run some validation locally, and I explained why below

## Manual Test Plan / Notes

No live relay or production mutation was needed.
The regression hides an event across two ticks, reveals it after the checkpoint advances, and verifies one Blossom DELETE.
Additional regressions verify overlap replay waits for transient retry backoff, does not inflate processed count for pure replay, still processes non-transient sibling targets, and rejects a kind-5 relay stream that closes before EOSE.
Delays beyond five minutes can still be missed; a stronger bound needs a relay indexing-lag guarantee.

## Visuals / API Examples

There is no visual or externally visible API change.

- [x] No visual change
- [ ] Screenshots, sample payloads, or logs attached if needed

## Public Artifact Review

The public text uses only repository and protocol terms needed to explain the fix.

- [x] Titles, descriptions, branch names, and screenshots avoid partner, customer, brand, campaign, or other sensitive external names unless explicitly approved

### DIFF
--- a/src/creator-delete/cron.mjs
+++ b/src/creator-delete/cron.mjs
@@ -11,6 +11,10 @@ const LAST_POLL_KEY = 'creator-delete-cron:last-poll';
 const DEFAULT_LOOKBACK_SECONDS = 3600; // first run
 const POLL_OVERLAP_SECONDS = 300;
 
+function countProcessedTargets(result) {
+  return (result.targets || []).filter(target => !target.skipped).length;
+}
+
 export async function runCreatorDeleteCron(deps) {
   const { db, kv, queryKind5Since, fetchTargetEvent, callBlossomDelete, now = () => Date.now() } = deps;
   const nowMs = now();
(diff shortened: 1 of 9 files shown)


NotThatKindOfDrLiz (approved):
Takeover changes look ready: the delayed kind-5 recovery now preserves per-target retry backoff, strict relay polling prevents cursor advancement on truncated streams, replay no longer inflates processed counts, and the new regressions cover the loss cases. CI is green.

PR #192, opened 2026-08-06
fix(relay-admin): preserve refusal code/caseId/state on a blocked un-ban

### LINKED ISSUE #191: un-ban refusals lose code/caseId/state, so an age-review block reads as a generic failure
## Problem

`callRelayAdminAction` collapses a structured relay-manager refusal into a bare `Error`:

```js
if (!response.ok || data?.success === false) {
  throw new Error(data?.error || `Relay admin error: HTTP ${response.status}`);
}
```

Only `data.error` survives. `code`, `caseId` and `state` are discarded before any caller can route on them.

That did not matter while relay-manager answered `unbanpubkey` with a plain success or failure. divinevideo/divine-relay-manager#217 changes it: an un-ban of an account with an open age-review case now returns `409 age_review_active` carrying `caseId` and `state`, and `503 age_review_check_failed` when the check itself cannot run.

This service reaches that path (`src/index.mjs`, `unbanpubkey` at ~`:761`, called at ~`:874` and ~`:2019`), so once #217 lands a moderator un-banning an uploader under age review from this dashboard gets a generic failure with no indication that a case exists, no case id to open, and no way to tell a permanent refusal from a transient one.

## Why it matters beyond the message

The refusal exists to route the moderator to the case rather than dead-end them. Relay-manager's own UI does exactly that. Losing the code here means this surface gets the enforcement block without the routing that makes it actionable, which reads as the tool being broken.

## Expected

Preserve `code` (and `caseId`/`state` where present) on a refusal so callers can distinguish:

- `age_review_active` — permanent, route to the case
- `age_review_check_failed` — the check could not run, retryable
- anything else — a genuine relay failure

## Notes

Found by @realmeylisdev while reviewing divinevideo/divine-relay-manager#217. Not a defect in that PR; it is a consequence of landing it, and worth closing before or shortly after it ships.

Related: divinevideo/coop-webhook-adapter#6 covers the same class of loss on the COOP enforcement adapter path. This one is separate.

realmeylisdev:
## Summary

- `callRelayAdminAction` now throws a `RelayAdminError` carrying `status`/`code`/`caseId`/`state` instead of flattening every relay-admin failure into `new Error(data.error)`.
- `POST /admin/api/uploader/:pubkey/enforcement` maps that onto a status a caller can route on: **409** for `age_review_active` (with `caseId`, `state`, and a `caseUrl` deep-link), **503** for `age_review_check_failed`, **502** for everything else exactly as before.
- The dashboard shows a refused un-ban with an **Open case** action instead of a dead-end red toast. `showToast` gained an `actionLabel` parameter; the default is still `Undo`, so no existing call site moves.
- Documented the cross-service contract in `docs/integrations/` and `RELAY_ADMIN_UI_URL` in `wrangler.toml`.

## Motivation

`divinevideo/divine-relay-manager#217` merged, so an un-ban of an account with an open age-review case now returns `409 age_review_active` carrying the case, and `503 age_review_check_failed` when the check itself cannot run. This service maps `allow_pubkey` onto `unbanpubkey`, so it reaches that path — and discarded everything but the message.

Both refusals therefore arrived as an identical flat 502. A moderator un-banning an uploader under age review got a generic failure with no case to open and no way to tell a permanent refusal from a transient one. The refusal exists to route the moderator to the case rather than dead-end them; without the code, this surface got the enforcement block without the routing that makes it actionable.

Scope note: only `allow_pubkey` (`unbanpubkey`) is guarded upstream. `ban_pubkey` and `delete_event` map to `banpubkey`/`banevent`, neither of which the age-review guard covers, so those paths are unchanged. The issue cites a second call site around `:874`, but that one only ever sends `delete_event`.

## Related Issue

- Closes #191

## Validation

- [x] `npm run lint` — passes (180 module files, 4 HTML files)
- [x] `npm test` — 1470 passing, 82 files
- [x] relevant Wrangler or queue-flow validation if applicable
- [ ] I could not run some validation locally, and I explained why below

## Manual Test Plan / Notes

New endpoint tests in `src/uploader-enforcement.test.mjs` stub relay-admin with each real response shape and assert the mapping, that the guarded method was `unbanpubkey`, and that the local `uploader_enforcement` row still records the ban the relay never lifted. Each was mutation-checked before being kept (counts below are on the current head, `ab9671e`, and will drift as tests are added — re-run the mutation rather than trust the number):

| Mutation | Result |
|---|---|
| throw a plain `Error` again (drop the structured fields) | reds 7 tests |
| disable the `age_review_check_failed` branch | reds the 503 test |
| build the case link from the API host instead of the UI host | reds the `RELAY_ADMIN_UI_URL` test |
| let anything but the 409 refusal fall through and write the enforcement row anyway | reds 5 tests |

`dashboard.html` has no build step, so its own tests substring-match the source — and substring matching can't tell a real fix from a dead one: reducing the throw site to `throw new Error(failure.message)` leaves `error.caseUrl` undefined, so the **Open case** branch never runs and every 409 shows the dead-end toast this PR exists to remove, while every substring assertion stays green. `src/admin/dashboard-enforcement.test.mjs` closes that gap: it extracts `showToast`, `hideToast`, and `updateUploaderEnforcement` out of the page source with a brace-aware scanner (immune to reformatting, and pinned to a line only the real function body contains, so a stale doc comment can't hijack the extraction), then drives them against stubs for each response shape — 409 with a usable case, 409 with no usable case (non-https link or a blank/non-string `caseId`), 503, 502, and success including a refresh that itself throws. Mutation-checked against dropping the structured throw, either carried field, the https gate, or the refresh-error isolation guard.

Not validated on staging — the relay-manager side is already on `main`, and this change is only reachable through a real age-review case.

## Visuals / API Examples

- [ ] No visual change
- [x] Screenshots, sample payloads, or logs attached if needed

`POST /admin/api/uploader/<pubkey>/enforcement` with `{"relayBanned": false}`, refused:

```json
HTTP 409
{
  "success": false,
  "error": "This account is under age review. Restrict or clear it from the Age Review flow.",
  "code": "age_review_active",
  "caseId": "2f3a1c48-9d5e-4b17-9c0a-6e8b1d7f4a20",
  "state": "restricted_pending_parental_consent",
  "caseUrl": "https://relay.admin.divine.video/age-review?case=2f3a1c48-9d5e-4b17-9c0a-6e8b1d7f4a20"
}
```

```json
HTTP 503
{
  "success": false,
  "error": "Could not check age-review status. Try again.",
  "code": "age_review_check_failed"
}
```

Any other relay failure keeps the previous shape: `HTTP 502 {"success": false, "error": "<message>"}`.

UI change is one branch in the existing toast: same component, same styling, with the action button labelled `Open case` instead of `Undo`.

## Public Artifact Review

- [x] Titles, descriptions, branch names, and screenshots avoid partner, customer, brand, campaign, or other sensitive external names unless explicitly approved


hm21 (approved):
LGTM. The server side is right and genuinely tested — I verified the upstream contract against `divine-relay-manager` (`ageReviewActiveGuard` really does answer `409 age_review_active` + `caseId`/`state` and `503 age_review_check_failed`), confirmed the scope note (the second `callRelayAdminAction` call site only ever sends `delete_event`), and reproduced your mutation table: swapping `RelayAdminError` back to a plain `Error` reds exactly 3 tests.

Pushed three mechanical fixes as separate commits — revert any you disagree with:

- `3242d2d` — the comment at `dashboard.html:3483` claimed the fallback toast "still carries the refusal and the case id". It doesn't: `blockedMessage` is scoped inside `if (caseUrl)`, so the fallback interpolates `error.message` only.
- `0c95b3b` — `failure.code` was assigned and never read anywhere (the catch routes on `caseUrl`/`caseId`, consumes the error, never rethrows). Dropped it and the test line that pinned it.
- `47a6212` — doc said "the only caller"; there are two. Also qualified the **Open case** action with the https gate.

`npm run lint` clean, 1457/1457 passing on the result.

One thing worth your call, not pushed:

**The dashboard half has no test that can fail for the right reason.** `throw failure;` (`dashboard.html:3452`) is the single line the whole feature hangs on. Mutate it to `throw new Error(failure.message);` and the feature is completely dead — `error.caseUrl` is `undefined`, `if (caseUrl)` never runs, every 409 shows the dead-end toast this PR exists to remove — and the suite stays **green**. Meanwhile splitting the `showToast(...)` call across five lines with byte-identical behaviour turns it **red**. The signal is inverted.

You already built the fix and threw it away: *"extracted into a throwaway harness and run against all five response shapes"*. That harness is what should have been committed. It's feasible in this runtime with no build step and no new dependency — regex the throw site and the catch site out of the imported HTML, `new Function` them, drive them against stubs. That catches the `throw failure;` mutation and is immune to reformatting.

Smaller, same class: `relayAdminField`'s type guard and `encodeURIComponent(caseId)` have zero coverage — I gutted both and the full 1457-test suite passed.

CI red is not yours: all three jobs died at *"Prepare all required actions / Service Unavailable"*. Re-running.

NotThatKindOfDrLiz (approved):
Approved on the current head 48de261 after takeover.

Final pass is clean. The only takeover change is the one-line assertion that the 409 age-review Open case toast keeps the intended 10000ms duration, retiring the unused timer scaffolding called out in review. Local lint/tests and GitHub CI/Semantic PR are green on head.

mbradley (approved):
Reviewer takeover on `48de261`, 14 commits pushed. The server half is right and the upstream contract checks out: I read `divine-relay-manager` at `origin/main` (`ce2c337`) and confirmed the 409 body shape, `age_review_check_failed`, fail-closed on reversals only, `banpubkey` deliberately unguarded, and the terminal states. Everything the new doc says about upstream is accurate.

@hm21's call on the dashboard tests was the right one, and the harness answers it. I re-ran the mutation: `throw failure;` to `throw new Error(failure.message);` now reds 2 tests where it used to stay green.

## Two coverage gaps, both closed

**The "no local row on a refusal" invariant was only asserted on the 409 path.** Keep the 409 returning early and let just 503 and 502 fall through to `setUploaderEnforcement`, and the whole suite stays green. That is the wrong path to leave uncovered: a permanent refusal failing this way is loud, because the moderator is already being sent to a case, but a transient one silently marks an account un-banned that the relay is still holding. Same assertion the 409 test already makes, added to the other two paths.

**The refresh-error isolation guard had no test.** Deleting the inner `try/catch` in `updateUploaderEnforcement` left every dashboard test green, because the stubs never rejected. Its comment says a reload hiccup must not look like the action failing, and the outer catch says "User action failed", so the regression tells a moderator to redo an enforcement that already landed. The harness now drives a rejecting refresh.

## Also pushed

- `relayAdminField` tested `value.trim()` but returned `value`, so a padded `caseId` became `?case=%20...%20`: a live Open case button resolving to no case.
- `error` was the one field read straight off the body while `code`/`caseId`/`state` were type-guarded. A non-string one reached the toast as `[object Object]`.
- `RelayAdminError.status` is the status received from relay-admin, but `error.status` in this same file means the status to respond with (`createHttpError`, echoed back by two handlers as `jsonResponse(error.status || 500, ...)`). Renamed to `upstreamStatus`. Latent, not live.
- `toast-undo` renamed to `toast-action`, finishing the rename that took `onUndo` to `onAction`. Left `swipe-review.html` alone; its button really is only ever Undo.
- The no-case-id test paired `caseId: null` with a populated `caseUrl`, which the worker cannot emit, since `caseUrl` is built only when the id survived. It now asserts the reachable shape.
- `extractFunction` guards against the one silent failure it had: the brace scanner is mode-aware but the name search is a plain regex over raw source, so an earlier textual `function showToast(` in a comment hijacks it and yields a phantom that still parses. Each extraction now names a line only the real body contains.
- Five doc corrections: what the 503 toast actually shows, every trigger for the plain-toast fallback, "not retryable" rather than "permanent", the `?pubkey=` lookup-cost claim (`AgeReview.tsx` resolves from the loaded list first, and `?case=` pays its own fallback fetch on the same terms), and the transport failures that throw before any body exists.

## Validation

`npm run lint` clean. `npm test`: 82 files, 1470 passing, up from 1467. Each fix was mutation-checked: reverted, watched the new test go red, restored. CI green on `bbc4e19`.

## Not pushed, for you to decide

**1. `ageReviewCaseUrl` builds the link by string concatenation.** A trailing slash on `RELAY_ADMIN_UI_URL` yields `https://host//age-review?case=...`, which passes the `https:` gate and dead-ends, so it degrades worse than the non-https case you explicitly handle. `new URL('/age-review', base)` plus `searchParams.set` normalizes it. Left to you because it feeds a security control, and `PR_REVIEW.md` puts that outside what takeover should touch without a separate say-so.

**2. A refused un-ban leaves `relay_banned = 1` locally, and that flag forces `PERMANENT_BAN` on every subsequent upload** (`applyUploaderEnforcementToResult`). `setUploaderEnforcement` has exactly one caller, reachable only past the relay call, so there is no local-only way to clear it. A moderator who concludes an earlier ban was wrong cannot undo it for the life of the case. Conservative may well be the right posture while a minor-safety case is open, but right now it is incidental rather than chosen. Options: leave it and say so in the doc, or add a local-only clear that does not call the relay. Which?

**3. The refusal toast is the only failure path in the handler without the "User action failed:" prefix**, and `.toast` has a single neutral style with no error variant, so it is pixel-identical to the success toast. The only failure cue is the red `setLookupStatus` line. A moderator who reads it as a successful hand-off believes the un-ban landed and nobody follows up. Options: prefix it like the others, add an error variant to `.toast`, or leave it on the grounds that the message text carries it.

**4. 503 and 502 render identically.** The service distinguishes retryable from permanent, and `0c95b3b` removed `code` from the client, so the only thing separating them is relay-admin's message happening to say "Try again." Nothing here tests that it keeps saying it. Options: route on `code` and give the 503 a Retry action, or accept the coupling and note it.

**5. `state` is computed, serialized, tested and documented, and has zero consumers.** There is no `data.state` in the dashboard. The states are not interchangeable: `open_reported` on a weeks-old hold is an escalation, `restricted_pending_parental_consent` is someone waiting on a third party who may never respond, `needs_follow_up` is stalled. Display it or drop it from the response.

**6. Is the extraction harness the go-forward pattern for dashboard tests, or a one-off for this path?** It is a real improvement over substring matching and worth keeping either way, but nothing outside the test file records the choice.

## Two things in the PR description

The mutation table says "reds 3 tests" for both the plain-`Error` row and the fall-through row. Measured on `48de261`, both red 6: the last three came in with `8ce07b6`, after the table was written. And the paragraph describing a "throwaway harness" now describes a committed 374-line file, and the coverage it claims from substring matching was removed by `21c8a1b`. Left alone since it is your prose.

## Remaining uncertainty

The fail-closed direction is correct, and I want that on the record given items 2 and 4 could read as objections to it. Failing open would silently lift a minor-safety hold while reporting success, which is strictly worse. One thing worth knowing rather than fixing: during a relay-side D1 outage every ban succeeds and every un-ban 503s, and there is no bound on that window, no alert (`wrangler.toml` sets no logpush or tail consumers), and no retry. Upstream's stated control is that such an outage has to be fixed rather than papered over. That is a platform question, not a question about this PR.

Over to you: review the takeover commits, revert any you disagree with, and merge when the six questions above are settled.

PR #193, opened 2026-08-07
fix(dm): classify moderation report DMs from NIP-17 tags and repair the dead unread badge

realmeylisdev:
## Summary

Fixes both halves of divine-mobile#6593 ("moderation DM intake is half-wired and its unread badge is dead code"):

1. **`dm-reader.mjs` only ever promoted a DM to a `user_reports` row when its content parsed as JSON `{type:'conversation_report', sha256}`** -- a shape no client, in any repo, has ever produced since this shipped (PR #24, four days before divine-mobile's report DM even existed). Reworked the classify step to read `sha256`/`report_type` from NIP-17 tags on the rumor instead -- content stays plain human-readable prose (matching NIP-17's "content MUST be plain text" convention and requiring no admin-UI render change), tags are the machine-readable side. divine-mobile's companion PR attaches these tags via `dm_repository.sendMessage`'s existing `additionalTags` parameter.

   Content-report DMs with a valid `sha256` now use the shared `recordReportForReview()` ingestion path, matching the HTTP and relay-report writers: valid report DMs create the `user_reports` row plus the corresponding `moderation_results` review/age-restriction state and AI-report telemetry. Invalid hash tags are ignored for report persistence rather than creating orphan rows.

   `user_reports.sha256` is `NOT NULL` with no user-only/message-only shape, so this can only activate for content reports with a resolvable video hash -- user reports and DM-message reports get a correctly-classified `dm_log` row but not a `user_reports` row, which is intentional (see the linked plan's "Non-goals"), not a gap this PR leaves open.

   Also extracts the per-event classify/persist logic into a separately exported `processRumor()` so it's directly testable against a plain rumor object -- this code path had **zero** test coverage before this change.

2. **The admin Messages UI's unread badge/dot read `conv.unread || conv.has_unread`, a field `getConversations()` never produced** -- exhaustive grep found zero producers anywhere in the schema or query layer, in any environment, ever. Adds a `dm_conversation_read_state` table (one row per conversation), a real `unread` computation in `getConversations` keyed off the latest *incoming* message (a moderator's own reply never marks their own inbox unread), a `POST /admin/api/messages/:pubkey/read` endpoint, and wiring in both admin pages including a mark-read call when a conversation is opened.

   The read endpoint now validates both route shape and recipient pubkey before writing read state, so malformed `/admin/api/messages/.../read` paths cannot create junk conversation markers.

## Test plan

- [x] `npm test` -- 1475/1475 pass across 81 files (33 new: unread/read-state, processRumor classify/review-ingestion, invalid hash, and read-route validation coverage)
- [x] `npm run lint` -- passes for all module + HTML files
- [x] New `unread`/`read_at` SQL logic tested against real D1 (`cloudflare:test`), not a mock -- including the exact timestamp-format hazard the migration's own comment calls out (`read_at` and `created_at` must both come from SQLite `CURRENT_TIMESTAMP` or `>` comparisons silently break)

## Related

- divine-mobile#6593
- Companion PR in divine-mobile (attaches the NIP-17 tags this reads)

### DIFF
--- a/migrations/012-dm-read-state.sql
+++ b/migrations/012-dm-read-state.sql
@@ -1,0 +1,23 @@
+-- Per-conversation read marker for the admin Messages UI's unread badge.
+--
+-- A separate one-row-per-conversation table (rather than a `read_at` column
+-- on `dm_log` itself) because "read" is a conversation-level concept, not a
+-- per-message one -- there is nothing sensible for an individual message row
+-- to mean by "this one message is read" independent of the rest of its
+-- thread. `getConversations` (dm-store.mjs) LEFT JOINs this table by
+-- conversation_id; a missing row means "never marked read", which combined
+-- with any existing incoming message correctly reads as unread.
+--
+-- `read_at` is always written via SQL `CURRENT_TIMESTAMP` (see
+-- markConversationRead in dm-store.mjs), matching `dm_log.created_at`'s own
+-- generation mechanism exactly. Both columns therefore share the same
+-- SQLite CURRENT_TIMESTAMP text format ("YYYY-MM-DD HH:MM:SS", UTC, no
+-- timezone marker) and remain safely comparable with a plain `>` -- mixing
+-- that format with a client-generated ISO-8601 string (which sorts
+-- differently under a lexicographic TEXT comparison because of the 'T'/'Z'
+-- characters) would silently make "unread" permanently stick at whatever it
+-- was the first time a conversation was marked read.
+CREATE TABLE IF NOT EXISTS dm_conversation_read_state (
+  conversation_id TEXT PRIMARY KEY,
+  read_at TEXT NOT NULL
+);
--- a/src/admin/dashboard.html
+++ b/src/admin/dashboard.html
@@ -5452,7 +5452,7 @@
         if (res.ok) {
           const data = await res.json();
           const convs = data.conversations || data || [];
-          const unread = convs.filter(c => c.unread || c.has_unread).length;
+          const unread = convs.filter(c => c.unread).length;
           if (unread > 0) {
             const badge = document.getElementById('unread-badge');
             if (badge) {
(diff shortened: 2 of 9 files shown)


hm21 (approved):
LGTM once the two commits I pushed land with you. Core premise holds — org-wide code search confirms the JSON `conversation_report` shape has zero producers outside this repo, and the tag names match divinevideo/divine-mobile#6831.

Two findings, pushed as separate commits rather than left as comments:

**6cd4aaa** — `getConversations` LEFT JOINs `dm_conversation_read_state`, but CI's deploy job runs `wrangler deploy` on merge and never `d1 migrations apply`. Merging would have shipped a worker querying a table prod doesn't have until migration 012 was applied by hand, breaking the whole Messages list and the dashboard fetch in the gap — not just the badge. `initDmLogTable` created the table but has no production caller (tests only), so it didn't cover this. Extracted the DDL into `initDmReadStateTable` and ensured it on the two admin routes that touch the table, rather than the per-request block in the fetch handler (that one runs for API traffic too). Both new tests fail without the ensure.

**43c1cd2** — the mobile producer attaches `report_type` to all three report variants but `sha256` only to content reports, so keying `message_type` off `sha256` alone filed user reports and DM-message reports as `creator_reply` — indistinguishable from ordinary chat in the admin UI. Not creating a `user_reports` row for those is the stated non-goal and still gated on `sha256`; losing the badge looked like a separate, unintended gap. `badgeClass`/`badgeLabel` derive from the string, so no UI change needed.

Also **bc2766d**: replaced `void firstReadAt` in the read-state test with a real assertion.

Validated: `npm run lint` clean, `dm-reader` + `dm-store` + `index` suites 223/223 locally, CI Lint + Test green on the new head. Please revert or adjust anything you disagree with before merging.

NotThatKindOfDrLiz (approved):
Approved after takeover commit a1180c5. Final pass is clean: report DM persistence now uses the shared review ingestion path, invalid sha256 tags are ignored for report persistence, read-route junk writes are blocked, and local plus GitHub validation are green.

PR #194, opened 2026-08-07
fix(dm): classify report DMs from the NIP-32 label and decode the full label vocabulary

realmeylisdev:
Refs divinevideo/divine-mobile#6593 (backend half). Replaces #193, which this supersedes -- it carries #193's six approved commits unchanged, plus the cross-service fix that came out of review on the mobile side.

Companion: divinevideo/divine-mobile#6855. Neither half works alone; they should merge together.

## Carried over from #193 (unchanged, already approved)

- Classify moderation report DMs from NIP-17 tags instead of trying to `JSON.parse` the content -- no client has ever sent that JSON.
- Repair the dead unread badge: real `getConversations` unread computation, a read-state table, and a mark-read endpoint. `dashboard.html` and `messages.html` both read `c.unread || c.has_unread`, which nothing ever set.
- Badge a report DM from `report_type`, not only `sha256`, so user reports and DM-message reports don't render as ordinary replies.
- Record report DMs through the shared review ingestion path.

### 1. Resolve the DM's report type through the shared resolver

`processRumor` read `['report_type', x]` straight off the rumor while the kind-1984 poller ran `extractReportType`. Both write the same `(sha256, reporter_pubkey)` row under `INSERT OR IGNORE`, so two vocabularies meant whichever path ingested first pinned its answer permanently -- and the DM's NIP-56 value is the lossy one, collapsing `aiGenerated` to `other`.

`processRumor` now calls `extractReportType` on the rumor. `extractReportType` gains a `report_type` arm, placed **above** the `Reason:` content regex: a DM's content is localized prose (`Reason: Contenu indésirable`), so letting the regex outrank an explicit tag would make `report_type` locale-dependent. kind-1984 events carry no `report_type` tag, so that arm is inert for them and their behaviour is byte-identical.

### 2. Decode the full social.nos.ontology label vocabulary

`extractReportType` prefers the NIP-32 `l` label, and divine-mobile sets one on every report. `normalizeReportType` lowercases and maps `-` to `_`, so `NS-sexualContent` arrived as `ns_sexualcontent` -- a value that matches nothing in `NSFW_REPORT_TYPES`. Only the `aiGenerated` family was special-cased.

The fix resolves each `social.nos.ontology` label to its canonical type, mirroring the table `divine-relay-manager` already keeps in `CATEGORY_LABELS`. That corrects stored `report_type`, sets adult categories for sexual-content reports, and lets AI telemetry fire from both ingestion paths.

kind-1984 relay reports still pass `source: 'relay-report'`, so `recordReportForReview` leaves `allowAutoAgeRestrict` false for that path. This PR fixes relay report typing, but it does **not** make the relay path auto age-restrict. Section 4 records the matching decision for the DM path.

The mapping is guarded by tests for all eleven current labels, namespace filtering, and a warning when a future `NS-*` label is not mapped.

Tracked independently in #195.

### 3. Contract test with real crypto

The `processRumor` unit tests hand it a plain rumor object, so nothing covered the step where the tags must survive NIP-44 encryption and seal/gift-wrap serialization. `dm-report-contract.test.mjs` starts from the exact tag list divine-mobile builds, wraps and unwraps with real NIP-59 crypto and schnorr signing, and asserts on the rows the real classify path writes.

### 4. Report DMs are REVIEW-only (policy decision)

Review asked for an explicit call on whether an unauthenticated report DM should be able to auto age-restrict public content from a client-supplied `sha256`. **The decision is no.** `9e517d8` drops the `allowAutoAgeRestrict: true` override, so `source: 'dm-report'` takes the same REVIEW-only default that `'relay-report'` already takes.

The trust boundary this records: a report DM is the least verified report the service accepts. `rumor.pubkey` is whatever key reached the moderation inbox, the `sha256` is client-supplied, and unlike the relay poller this path does not fetch the target event, require a Divine client, or pass the processed-key gate. Auto-escalation is left to the authenticated `/api/v1/report` route, which was already the only path with that authority -- the relay poller has never had it, so this makes the two Nostr paths consistent rather than adding a new restriction.

The DM still writes its `user_reports` row and is still badged `conversation_report` in the admin Messages UI. Two NSFW report DMs surface for review exactly as before; they just no longer hide public content without a human. Pinned by a test rather than a comment -- `dm-report-contract.test.mjs` drives two distinct reporter keys through the real gift-wrap round trip and asserts the result stays `REVIEW`.

### 5. Close the same boundary from the other side

Keeping the DM path REVIEW-only stops a report DM escalating *on its own*, but not as one of two. `addReport` counted `DISTINCT reporter_pubkey` per `sha256` with no source filter, and `POST /api/v1/report` still auto age-restricts at two distinct reporters -- so a minted key sending one report DM plus one genuine authenticated report was enough to hide public content, with the unverified half free.

Nothing produced that shape before this branch: the DM path's predecessor only wrote a `user_reports` row from inside the `JSON.parse(content)` branch looking for `type: 'conversation_report'`, the JSON no client has ever sent. Tag-based classification is what makes DM-sourced rows real, so the threshold fix belongs here.

`user_reports` gains a `source` column. `distinctReporterCount` keeps its meaning (every reporter, any source -- the admin UI and the `/api/v1/report` response both surface it); a new `escalationReporterCount` excludes `dm-report` rows and is what gates `AGE_RESTRICTED`. Both are written into `moderation_results.raw_response` so a moderator can see why something did not escalate. Rows predating the column are NULL and still count, since none of them can be DM-sourced.

The column is added by `initReportsTable` rather than a file in `migrations/`, matching how `user_reports` is created in the first place, with an idempotent `ALTER` for existing databases. `syncInbox` and `pollRelayForReports` call `initReportsTable` too: both run from the cron trigger, which never reaches the fetch handler's schema ensure.

### 6. Stop trusting a report DM's self-reported timestamp

`rumor.created_at` is written by the sender and validated by nothing, and it flowed straight into `new Date(createdAt * 1000).toISOString()`. A missing or non-numeric value threw `RangeError` inside `processRumor`'s warn-and-continue block, discarding the whole report -- no `user_reports` row, no `REVIEW` row, while the DM still landed in `dm_log` and looked handled. A backdated value landed in `moderation_results.moderated_at`, which the admin queue sorts on and its `since` filter drops outright.

`resolveReportedAt` uses the rumor's timestamp only where it could plausibly be true -- no older than the reader's own first-run lookback, no further ahead than a clock-skew allowance -- and falls back to receipt time otherwise, so both failure directions round toward "surfaces now" rather than "surfaces never".

### 7. A reporter's escalation eligibility no longer depends on arrival order

`user_reports` is keyed `UNIQUE(sha256, reporter_pubkey)`, and one report reaches two ingestion paths: divine-mobile publishes the kind-1984 **and** sends the report DM for the same content from the same key. Section 3's contract test already pins that pairing -- its last case asserts the two paths agree on `report_type`. They cannot agree on `source`, which names the path they arrived through, so `INSERT OR IGNORE` left arrival order deciding whether that reporter counted toward `escalationReporterCount` -- permanently, once pinned.

Within one cron tick `pollRelayForReports` runs before `syncInbox`, so the ordinary mobile case pins `relay-report` and counts. The mis-pin needs the DM to land without a successful relay ingest first: a kind-1984 publish that failed while the gift wrap landed, the `RELAY_REPORTS_REQUIRE_DIVINE_CLIENT` gate, poller lookback/pagination, or the dialog's resubmit path.

`addReport` now upgrades `source` on conflict, in one direction only -- `dm-report` to any escalation-eligible source. A later report DM never downgrades a reporter the HTTP or relay path established, so the untrusted channel cannot promote itself; it can only fail to be promoted. `report_type`, `reason` and `created_at` stay first-write-wins, so the report of record is still the one first filed.

This is the one change in this PR that loosens rather than tightens: a genuine reporter who happened to DM first counts again toward the authenticated path's two-reporter threshold.

### 8. The report API stops returning a verdict it did not make

`POST /api/v1/report` returned `escalate` from `addReport`'s own 3-and-5-distinct-reporter thresholds, which nothing in the service enforces. The real decision is made a line later in `recordReportForReview`, on different inputs (escalation-eligible reporters, not every reporter) and different thresholds (2, and only for NSFW types) -- so the two could disagree outright. Five `spam` reports wrote `REVIEW` to `moderation_results` while telling the client `AGE_RESTRICTED`, and section 5's `source` column widened the gap by letting DM rows inflate the count the response used but not the one enforcement used.

Deriving the guess from `escalationReporterCount` would close only the second half, so the computation is deleted instead: `addReport` counts reporters and cannot see report type or source policy, so it was never in a position to have an opinion. `escalate` now carries the decision this report was recorded under -- the same value the moderation row is written from -- which makes contradicting enforcement impossible by construction. (Not a claim about the row's final state: `recordReportForReview`'s `ON CONFLICT` keeps a stronger prior action, so a sha256 already `PERMANENT_BAN` or human-reviewed stays that way.)

| reporters (authenticated path) | recorded action | `escalate` before | after |
|---|---|---|---|
| 1 x nudity | `REVIEW` | `null` | `"REVIEW"` |
| 2 x nudity | `AGE_RESTRICTED` | `null` | `"AGE_RESTRICTED"` |
| 5 x spam | `REVIEW` | `"AGE_RESTRICTED"` | `"REVIEW"` |
| 4 x DM + 1 x nudity | `REVIEW` | `"AGE_RESTRICTED"` | `"REVIEW"` |

Response shape is unchanged. The value is no longer null -- a single report reads `REVIEW`, which is what the service did with it. No caller in any Divine repo reads this endpoint today.

### 9. Each side of a report DM dedupes against its own row

`syncInbox` re-processes the same gift wrap for two days, so `9e9b94d` added a `dm_log` lookup that returns before report ingestion. That gate is keyed on the wrong row. `processRumor` records the report inside a warn-and-continue block and calls `logDm` either way, so a transient D1 error on the `user_reports` insert still logs the gift wrap -- and the next tick then skips before it can retry. The report is gone permanently, counted as `deduped` in the sync log, while the DM sits in the admin Messages UI badged as a report. That is the failure shape section 6 exists to remove, reached by a different route.

A `dm_log` row proves the DM was stored, not that its report was. So each side dedupes against its own row: report ingestion skips once this reporter has both a `user_reports` row and a `moderation_results` review row for this sha256, and the `'skipped'` return happens once the gift wrap is in `dm_log`.

This keeps what the gate was for. `recordReportForReview` is not idempotent for a report whose timestamp fell back to receipt time -- `moderated_at` moves and the AI telemetry `event_key` (`report:<sha>:<type>:<createdAt>`) changes, so its `INSERT OR IGNORE` stops deduping -- and the completed report+review check suppresses exactly that. `9e9b94d`'s regression test, which pins that `moderated_at` is not re-bumped, passes unchanged. The `'skipped'` return also keeps `9e9b94d`'s accounting fix: `logDm` hands back the existing row on a dedup hit, whose truthy `.id` used to read as a fresh insert.

## Test plan

- [x] `npm run lint`
- [x] `npm test` -- 1539/1539 across 83 files
- [x] `npx vitest run src/reports.test.mjs src/index.test.mjs src/moderation/report-review.test.mjs src/nostr/dm-report-contract.test.mjs src/nostr/dm-reader.test.mjs src/nostr/report-poller.test.mjs` -- 293/293
- [x] Each new guard red-checked by reverting it: restoring `allowAutoAgeRestrict: true`, gating on `distinctReporterCount`, inlining the raw `created_at`, dropping the `ON CONFLICT` source upgrade (`expected +0 to be 1`), restoring the old `escalate` field (`expected 'AGE_RESTRICTED' to be 'REVIEW'`), and re-keying section 9's skip on only the `dm_log` row or only the `user_reports` row each turn their tests red.

### DIFF
--- a/src/nostr/dm-reader.mjs
+++ b/src/nostr/dm-reader.mjs
@@ -51,6 +51,12 @@ export function resolveReportedAt(createdAt, nowMs = Date.now()) {
   return new Date(isUsable ? createdAt * 1000 : nowMs).toISOString();
 }
 
+async function findModerationResultBySha(db, sha256) {
+  return db.prepare(
+    'SELECT sha256 FROM moderation_results WHERE sha256 = ?'
+  ).bind(sha256).first();
+}
+
 /**
  * Derive the moderator's hex pubkey from NOSTR_PRIVATE_KEY (hex)
  * @param {Object} env - Environment with NOSTR_PRIVATE_KEY
@@ -277,15 +283,18 @@ export async function processRumor(rumor, giftWrapId, moderatorPubkey, env) {
   // is what the relay path (`'relay-report'`) already takes. Only the
   // authenticated HTTP report API still auto-escalates.
   //
-  // On a re-poll, skip only once this reporter's row for this sha256 exists.
+  // On a re-poll, skip only once this reporter's row and the review row both
+  // exist for this sha256.
   // recordReportForReview is not idempotent for a report whose timestamp fell
   // back to receipt time: moderated_at moves and the AI telemetry event_key
   // (report:<sha>:<type>:<createdAt>) changes, so INSERT OR IGNORE stops
-  // deduping it. Keying the skip on the report row rather than the dm_log row
-  // suppresses that without also suppressing the retry a failed write needs.
+  // deduping it. Keying the skip on the completed report+review pair, rather
+  // than the dm_log row alone or user_reports alone, suppresses that without
+  // also suppressing the retry a partial failed write needs.
   if (relatedSha256 && env.BLOSSOM_DB && !isOutgoing) {
     const alreadyRecorded = alreadyLogged
-      && await findReportByReporter(env.BLOSSOM_DB, relatedSha256, senderPubkey);
+      && await findReportByReporter(env.BLOSSOM_DB, relatedSha256, senderPubkey)
+      && await findModerationResultBySha(env.BLOSSOM_DB, relatedSha256);
     if (!alreadyRecorded) {
       try {
         await recordReportForReview(env.BLOSSOM_DB, {
--- a/src/nostr/dm-reader.mjs
+++ b/src/nostr/dm-reader.mjs
@@ -208,6 +208,14 @@ export async function processRumor(rumor, giftWrapId, moderatorPubkey, env) {
   // present. Skip for outgoing self-copies: the reporter is the
   // counterparty, not the moderator (who is echoing their own sent
   // message).
+  // No allowAutoAgeRestrict override: a report DM is the least verified
+  // report we accept. The sender is any key that can reach the moderation
+  // inbox, the sha256 is client-supplied, and unlike the relay poller this
+  // path does not fetch the target event, require a Divine client, or pass
+  // the processed-key gate. So it records the report and leaves the call to
+  // a human -- `source: 'dm-report'` takes the REVIEW-only default, which
+  // is what the relay path (`'relay-report'`) already takes. Only the
+  // authenticated HTTP report API still auto-escalates.
   if (relatedSha256 && env.BLOSSOM_DB && !isOutgoing) {
     try {
       await recordReportForReview(env.BLOSSOM_DB, {
@@ -217,7 +225,6 @@ export async function processRumor(rumor, giftWrapId, moderatorPubkey, env) {
         reason: content,
         source: 'dm-report',
         reportedAt: new Date(createdAt * 1000).toISOString(),
-        allowAutoAgeRestrict: true,
       });
     } catch (reportErr) {
       console.warn(`[DM-READER] Failed to record report for review:`, reportErr.message);
--- a/src/nostr/dm-reader.mjs
+++ b/src/nostr/dm-reader.mjs
@@ -7,15 +7,15 @@
 import { getPublicKey } from 'nostr-tools/pure';
 import { hexToBytes } from '@noble/hashes/utils';
 import { unwrapEvent } from 'nostr-tools/nip17';
-import { computeConversationId, logDm } from './dm-store.mjs';
+import { computeConversationId, findDmByNostrEventId, logDm } from './dm-store.mjs';
 import { recordReportForReview } from '../moderation/report-review.mjs';
 import { extractReportType } from './report-poller.mjs';
 import { initReportsTable } from '../reports.mjs';
 import { isValidSha256 } from '../validation.mjs';
 
-// How far back a first sync looks for gift wraps. Also the floor for a
-// report's self-reported timestamp: nothing this reader ingests can honestly
-// be older than the oldest thing it fetches.
+// How far back a first sync looks for gift wraps. Also the plausibility floor
+// for a report's self-reported timestamp: older values are still ingested, but
+// are stamped with receipt time so they surface in the current review queue.
 const INBOX_FIRST_RUN_LOOKBACK_SECONDS = 7 * 86400;
 
 // Tolerance for a reporting client whose clock runs a little fast.
@@ -218,6 +218,15 @@ export async function processRumor(rumor, giftWrapId, moderatorPubkey, env) {
   // Compute conversation ID against the non-moderator side so inbound
   // and outbound messages in the same thread share a conversation_id.
   const conversationId = computeConversationId(moderatorPubkey, counterpartyPubkey);
+
+  // syncInbox deliberately overlaps two days of gift wraps on every tick for
+  // NIP-17 timestamp randomization. If this gift wrap already reached dm_log,
+  // stop before report ingestion so repeated polls cannot re-bump review rows
+  // or emit fresh AI-report telemetry. Keep logDm's own check too as a final
+  // write-side guard.
+  if (env.BLOSSOM_DB && await findDmByNostrEventId(env.BLOSSOM_DB, giftWrapId)) {
+    return 'skipped';
+  }
 
   // Structured report data travels as NIP-17 tags on the rumor, not as
   // JSON-encoded content -- content stays plain human-readable text
--- a/src/nostr/dm-reader.mjs
+++ b/src/nostr/dm-reader.mjs
@@ -10,6 +10,7 @@ import { unwrapEvent } from 'nostr-tools/nip17';
 import { computeConversationId, logDm } from './dm-store.mjs';
 import { recordReportForReview } from '../moderation/report-review.mjs';
 import { extractReportType } from './report-poller.mjs';
+import { initReportsTable } from '../reports.mjs';
 import { isValidSha256 } from '../validation.mjs';
 
 /**
@@ -39,6 +40,15 @@ export async function syncInbox(env) {
 
   const privateKey = hexToBytes(env.NOSTR_PRIVATE_KEY);
   const moderatorPubkey = getPublicKey(privateKey);
+
+  // The fetch handler ensures the reports schema on every request, but this
+  // runs from the cron trigger too, which never touches it. Without this, the
+  // first tick after a deploy could reach a user_reports table that has no
+  // `source` column yet and drop every report it decrypted -- processRumor
+  // catches the write failure and moves on.
+  if (env.BLOSSOM_DB) {
+    await initReportsTable(env.BLOSSOM_DB);
+  }
 
   // Get last sync timestamp from KV
   let lastSync = null;
--- a/src/nostr/dm-reader.mjs
+++ b/src/nostr/dm-reader.mjs
@@ -189,7 +189,10 @@ export async function processRumor(rumor, giftWrapId, moderatorPubkey, env) {
   const rumorTags = Array.isArray(rumor.tags) ? rumor.tags : [];
   const shaTag = rumorTags.find((t) => Array.isArray(t) && t[0] === 'sha256' && t[1]);
   const hasReportTag = rumorTags.some(
-    (t) => Array.isArray(t) && (t[0] === 'report_type' || t[0] === 'l') && t[1]
+    (t) => Array.isArray(t) && (
+      (t[0] === 'report_type' && t[1])
+      || (t[0] === 'l' && t[1] && t[2] === 'social.nos.ontology')
+    )
   );
   const rawSha256 = typeof shaTag?.[1] === 'string' ? shaTag[1].toLowerCase() : null;
   const relatedSha256 = rawSha256 && isValidSha256(rawSha256) ? rawSha256 : null;
--- a/src/nostr/dm-reader.mjs
+++ b/src/nostr/dm-reader.mjs
@@ -171,11 +171,12 @@ export async function processRumor(rumor, giftWrapId, moderatorPubkey, env) {
   // Structured report data travels as NIP-17 tags on the rumor, not as
   // JSON-encoded content -- content stays plain human-readable text
   // (matching NIP-17's "content MUST be plain text" convention) so the
-  // admin Messages UI can keep rendering it as-is. A report DM carries a
-  // ['sha256', <hash>] tag when the reported content has a resolvable
-  // Blossom blob hash (content reports only -- user reports and
-  // DM-message reports have no such hash and never carry this tag,
-  // by design: see divine-mobile#6593 plan's "Non-goals").
+  // admin Messages UI can keep rendering it as-is. Every report DM carries
+  // a ['report_type', <nip-56 type>] tag; only content reports also carry
+  // ['sha256', <hash>], because user reports and DM-message reports have no
+  // resolvable Blossom blob hash (by design: see divine-mobile#6593 plan's
+  // "Non-goals"). So report_type classifies the message, sha256 gates the
+  // user_reports row.
   const rumorTags = Array.isArray(rumor.tags) ? rumor.tags : [];
   const shaTag = rumorTags.find((t) => Array.isArray(t) && t[0] === 'sha256' && t[1]);
   const reportTypeTag = rumorTags.find((t) => Array.isArray(t) && t[0] === 'report_type' && t[1]);
@@ -205,8 +206,13 @@ export async function processRumor(rumor, giftWrapId, moderatorPubkey, env) {
     }
   }
 
-  // Log to dm_log (dedup by nostr_event_id)
-  const messageType = relatedSha256
+  // Log to dm_log (dedup by nostr_event_id). Either machine-readable tag
+  // marks the DM as a report: report_type is present on all three report
+  // variants, sha256 only on content reports. Keying the badge off sha256
+  // alone would render a user report or a DM-message report as an ordinary
+  // reply in the admin Messages UI, which is a display gap rather than the
+  // deliberate user_reports non-goal above.
+  const messageType = (relatedSha256 || reportTypeTag)
     ? 'conversation_report'
     : (isOutgoing ? 'moderator_reply' : 'creator_reply');
   const result = await logDm(env.BLOSSOM_DB, {
--- a/src/nostr/dm-reader.mjs
+++ b/src/nostr/dm-reader.mjs
@@ -10,7 +10,7 @@ import { unwrapEvent } from 'nostr-tools/nip17';
 import { computeConversationId, findDmByNostrEventId, logDm } from './dm-store.mjs';
 import { recordReportForReview } from '../moderation/report-review.mjs';
 import { extractReportType } from './report-poller.mjs';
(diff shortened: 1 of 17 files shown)


src/nostr/dm-reader.mjs
```diff
@@ -196,6 +113,140 @@ export async function syncInbox(env) {
...
+  if (relatedSha256 && env.BLOSSOM_DB && !isOutgoing) {
+    try {
+      await recordReportForReview(env.BLOSSOM_DB, {
+        sha256: relatedSha256,
+        reporterPubkey: senderPubkey,
+        reportType,
+        reason: content,
+        source: 'dm-report',
+        reportedAt: new Date(createdAt * 1000).toISOString(),
+        allowAutoAgeRestrict: true,
```
NotThatKindOfDrLiz: This is the remaining blocker: the DM path opts into auto age-restriction while trusting a `sha256` tag from the decrypted DM. Unlike the relay path, this path does not fetch/verify the target event, enforce Divine-client report origin, or use the relay processed-key gate. Please make the policy decision explicit before merge: either keep DM reports REVIEW-only, add equivalent trust checks, or document that unauthenticated report DMs are intentionally allowed to drive `AGE_RESTRICTED` after two distinct reporters.

NotThatKindOfDrLiz (changes requested):
Requesting changes for the remaining policy decision only. The bounded code issues from review have been addressed in takeover commit `44926e7`, and CI is green. Before merge, please make an explicit call on whether unauthenticated report DMs should be able to auto age-restrict public content from a client-supplied `sha256`. If the answer is no, remove this override or route DM reports through a weaker REVIEW-only path. If the answer is yes, document that trust boundary in the PR/body and issue so the moderation outcome authority is intentional.

src/nostr/dm-reader.mjs
realmeylisdev: Decision: **REVIEW-only.** Fixed in 9e517d8.

Dropped the `allowAutoAgeRestrict: true` override, so `source: 'dm-report'` now takes the
default in `recordReportForReview` — the same one `'relay-report'` already takes. Nothing
else moved: the DM still writes its `user_reports` row, still counts toward
`distinctReporterCount`, and is still badged `conversation_report` in the admin Messages
UI. Two NSFW report DMs surface for review exactly as before; they just no longer hide
public content without a human.

Your framing understated it slightly, which made the call easier. The DM path wasn't
merely weaker than the relay path — it was the *only* Nostr path that could auto-escalate
at all. `recordReportForReview` defaults `allowAutoAgeRestrict` to
`source === 'user-report'`, and `report-poller.mjs` passes `source: 'relay-report'`
without an override. So the verified path (fetches the target event, enforces Divine-client
origin, uses the processed-key gate) was already REVIEW-only, and the unverified one had
opted itself out of that. Removing the override makes the two consistent rather than
imposing a new restriction, and leaves auto-escalation only on the authenticated HTTP
report API.

Pinned by a test rather than a comment: the contract test now drives two distinct reporter
keys through the real gift-wrap round trip and asserts the result stays `REVIEW`. I put
the override back and confirmed it goes red (`expected 'AGE_RESTRICTED' to be 'REVIEW'`).

Also in this push: a0603e9 updates the fixture comment that named
`report_content_dialog.dart`'s `_reportDmTags()` — divinevideo/divine-mobile#6855 moved
that assembly to `ContentReportingService.moderationDmTags()` in response to your review
there. Fixture and behaviour unchanged.

1521/1521 local, checks green on 9e517d8.

src/reports.mjs
```diff
@@ -17,39 +17,79 @@ export async function initReportsTable(db) {
...
+ * @param {{sha256: string, reporter_pubkey: string, report_type: string, reason?: string, created_at?: string, source?: string}} report
+ * @returns {Promise<{escalate: 'AGE_RESTRICTED'|'REVIEW'|null, distinctReporterCount: number, escalationReporterCount: number}>}
  */
-export async function addReport(db, { sha256, reporter_pubkey, report_type, reason, created_at }) {
+export async function addReport(db, { sha256, reporter_pubkey, report_type, reason, created_at, source = null }) {
   await db.prepare(`
-    INSERT OR IGNORE INTO user_reports (sha256, reporter_pubkey, report_type, reason, created_at)
-    VALUES (?, ?, ?, ?, COALESCE(?, CURRENT_TIMESTAMP))
-  `).bind(sha256, reporter_pubkey, report_type, reason ?? null, created_at ?? null).run();
+    INSERT OR IGNORE INTO user_reports (sha256, reporter_pubkey, report_type, reason, created_at, source)
```
NotThatKindOfDrLiz, line 113: `INSERT OR IGNORE` still pins the first `source` for a `(sha256, reporter_pubkey)` pair. If this pubkey DMs first and later files an authenticated report for the same blob, the later `user-report` is ignored and the row remains `dm-report`, so that authenticated reporter never counts toward `escalationReporterCount`. This fails safe, but it is the same first-ingestion-pins-policy hazard this PR is closing for `report_type`. Please upgrade `source` on conflict when a later authenticated report replaces a prior `dm-report`, and add a focused regression test.

src/reports.mjs
```diff
@@ -17,39 +17,79 @@ export async function initReportsTable(db) {
...
     FROM user_reports
     WHERE sha256 = ?
-  `).bind(sha256).first();
+  `).bind(NON_ESCALATING_SOURCE, sha256).first();
 
   const count = row?.cnt ?? 0;
+  const escalationCount = row?.escalation_cnt ?? 0;
 
   let escalate = null;
   if (count >= 5) escalate = 'AGE_RESTRICTED';
```
NotThatKindOfDrLiz: This legacy `escalate` value still uses the unfiltered distinct count. `/api/v1/report` returns it directly, so the API can now say `escalate: "AGE_RESTRICTED"` even when the actual moderation action is `REVIEW` because a DM row supplied part of the count. Please derive this from `escalationReporterCount` or remove/rename it so the response cannot contradict the enforcement decision.

NotThatKindOfDrLiz (changes requested):
Thanks for resolving the original DM auto-age-restriction blocker. I am requesting changes for two remaining source/escalation consistency issues in `addReport` before merge. Both are small, but they sit on the same moderation-outcome boundary this PR is tightening.

src/reports.mjs
realmeylisdev: Fixed in `93d283b`.

Confirmed -- and it's reachable through a second door that matters more than the authenticated one. divine-mobile publishes the kind-1984 *and* sends the report DM for the same content from the same key, so both paths write this same `(sha256, reporter_pubkey)` row. The contract test added earlier in this PR already pins that pairing: it asserts the two paths agree on `report_type`. They cannot agree on `source`, because `source` names the path. So `INSERT OR IGNORE` let arrival order decide whether that reporter was escalation-eligible at all.

Within one cron tick `pollRelayForReports` runs before `syncInbox`, so the ordinary mobile case pins `relay-report` and counts. The mis-pin needs the DM to land without a successful relay ingest first -- a kind-1984 publish that failed while the gift wrap landed, the `RELAY_REPORTS_REQUIRE_DIVINE_CLIENT` gate, poller lookback/pagination, or the dialog's resubmit path. Once it happens it is permanent.

Fixed as an upgrade-on-conflict rather than a `user-report` special case, since `relay-report` needs the same treatment:

```sql
ON CONFLICT(sha256, reporter_pubkey) DO UPDATE SET source = excluded.source
  WHERE user_reports.source = 'dm-report'
    AND excluded.source IS NOT NULL
    AND excluded.source <> 'dm-report'
```

One direction only. A later report DM never downgrades a reporter the HTTP or relay path established, so the untrusted channel still cannot promote itself -- it can only fail to be promoted. `report_type`, `reason` and `created_at` stay first-write-wins, so the report of record is still the one first filed.

Worth naming plainly: the old behaviour failed safe, so this is the direction that changes an outcome. A genuine reporter who happened to DM first now counts again toward the two-reporter threshold.

Tests: four in `reports.test.mjs` (upgrade via `user-report`, upgrade via `relay-report`, no downgrade from a later DM, and that only `source` moves), plus an end-to-end case in `dm-report-contract.test.mjs` driving the same key through the real gift-wrap round trip and then `POST /api/v1/report`. Red-checked by reverting the `ON CONFLICT` clause -- three of the four unit tests and the contract test go red (`expected +0 to be 1`). The no-downgrade test is green either way by design; it guards against a future over-broad upgrade, not this bug.

src/reports.mjs
realmeylisdev: Fixed in `ee76b3d` -- though not by deriving it, because that leaves half the contradiction standing.

`escalate` and the enforcement decision disagreed on two axes. Inputs: every reporter vs escalation-eligible reporters -- the half you found. Thresholds: 3-and-5 for any report type vs 2 for NSFW only. Deriving `escalate` from `escalationReporterCount` closes the first, but five authenticated `spam` reports would still return `escalate: "AGE_RESTRICTED"` next to a row that says `REVIEW`. Same contradiction, different door.

So I removed the computation instead. `addReport` counts reporters; it cannot see the report type or the source policy, so it was never in a position to have an opinion -- and nothing in the service enforced its thresholds. `/api/v1/report` now returns the action that was actually recorded (`escalate: result.action`).

| reporters (authenticated path) | recorded action | `escalate` before | after |
|---|---|---|---|
| 1 x nudity | `REVIEW` | `null` | `"REVIEW"` |
| 2 x nudity | `AGE_RESTRICTED` | `null` | `"AGE_RESTRICTED"` |
| 5 x spam | `REVIEW` | `"AGE_RESTRICTED"` | `"REVIEW"` |
| 4 x DM + 1 x nudity | `REVIEW` | `"AGE_RESTRICTED"` | `"REVIEW"` |

Response shape is unchanged. The value is no longer null -- a single report reads `"REVIEW"`, which is what the service did with it. No caller reads this endpoint today; I checked every Divine repo, and the `/api/v1/report` hits outside this one belong to an unrelated third-party service. `support-trust-safety`'s vocabulary audit records the same.

Tests: `index.test.mjs` now asserts the response `escalate` equals the action bound into the `moderation_results` write on both existing endpoint cases, plus a new case for the 5-reporter non-NSFW shape. Red-checked by restoring the old field: `expected 'AGE_RESTRICTED' to be 'REVIEW'`.

NotThatKindOfDrLiz (approved):
Approved after takeover commit `9e9b94d`.

I verified the remaining blocker is fixed: repeated processing of the same gift wrap now stops before `recordReportForReview`, so the two-day NIP-17 overlap cannot re-bump review rows or mint duplicate AI-report telemetry. The existing `logDm` dedup remains as the write-side fallback, and the regression test now pins that `moderation_results.moderated_at` stays unchanged on a duplicate pass.

Local validation passed: targeted vitest 78/78, `npm run lint`, and full `npm test` 1537/1537. CI is green on the pushed head: Lint, Test, semantic PR, and CLA; Deploy skipped as expected.

Do not merge this independently of divinevideo/divine-mobile#6855; the PR body now uses a non-closing issue reference for that reason.

mbradley (approved):
Reviewed the delta since Dr Liz's approval — `2a30506` only, a clean child of `9e9b94d`, no rebase.

That commit reworks the dedup gate so a report whose write failed transiently is retried on the next pass instead of being permanently skipped once the gift wrap is in `dm_log`. Verified at head:

- `syncInbox` runs `initReportsTable` before the `processRumor` loop, so the new `findReportByReporter` can't hit a missing table in prod; the return value only feeds the `synced/skipped/errors` counters.
- Skipping `recordReportForReview` on a genuine re-poll still holds: `moderated_at` is not re-bumped and AI telemetry is not re-minted. `source` stays `'dm-report'` and remains excluded from `escalationReporterCount`, so the retry path gives the untrusted DM source no new escalation power — the retry only surfaces a report sooner and only as `REVIEW`.
- Mutation-checked the new retry test: re-keying the skip on the `dm_log` row turns it red (`expected [] to have a length of 1 but got +0`), while the `moderated_at` regression test stays green. The coverage is real.

Approving `2a30506` on top of the prior approval. Per the PR body, merge in lockstep with divinevideo/divine-mobile#6855, not independently.

NotThatKindOfDrLiz (approved):
Approved after takeover commit f094009.

I re-checked the remaining partial-write blocker at head: a report DM is now considered deduped only after both the reporter row and the moderation review row exist, so a transient moderation_results write failure remains retryable even when dm_log and user_reports already succeeded. The new regression test covers that path against real D1.

Local validation passed: targeted 293/293, npm run lint, and full npm test 1539/1539. GitHub checks are green on f094009: Lint, Test, semantic PR, and CLA; Deploy skipped as expected.

Still merge this with divinevideo/divine-mobile#6855, not independently.

PR #199, opened 2026-08-13
feat(dm): allow pinning outbound DM relays via DM_RELAY_URLS

mbradley:
## The problem

Every NIP-17 DM this Worker sends goes to `relay.divine.video` and three public relays, in every environment, and nothing in config can change that.

The publish targets come from module constants (`DIVINE_RELAY`, `DEFAULT_RELAYS`) rather than from `env`. So running this Worker anywhere other than production, with a real signing key, sends **real DMs to real users**. There is no setting that prevents it.

That is the same shape as pointing local dev at a production API, except the blast radius is a stranger's inbox rather than a database row.

## The change

`DM_RELAY_URLS`, when set, becomes the exact publish target list.

Setting it also skips the KV relay cache and kind-10002 discovery, because either of those would reintroduce a production relay and defeat the point. It also contains the kind-10050 inbox announcement: that event is published only when the configured inbox relay is in `DM_RELAY_URLS`, and only to that inbox relay. Unset is the production default and behaviour is unchanged.

A present value that parses to nothing is a configuration error. DM sends and the kind-10050 announcement fail closed instead of falling back to production relays. Only an absent `DM_RELAY_URLS` uses production defaults.

## Why it is worth having

It is what makes it safe to run this Worker outside production at all. Today the only safe way to exercise the DM paths locally is to not have a real signing key, which means the DM paths do not get exercised.

It also satisfies the `AGENTS.md` rule against hardcoding environment-specific domains in application code.

## Verification

- `npm run lint`
- `npm test`: 84 files, 1,586 tests passed

The containment tests assert on the transport targets and fail-closed paths rather than only on returned relay lists.

### DIFF
--- a/src/nostr/dm-sender.mjs
+++ b/src/nostr/dm-sender.mjs
@@ -333,15 +333,56 @@ async function recordRateLimit(recipientPubkey, env) {
 // --- Relay Discovery ---
 
 /**
+ * Parse the `DM_RELAY_URLS` publish-target override.
+ *
+ * Unset (the production default) returns [] and every caller keeps today's
+ * behaviour: discovered relays, DIVINE_RELAY, DEFAULT_RELAYS.
+ *
+ * Set, it pins outbound DMs to exactly these relays. That containment is what
+ * makes it safe to run this Worker outside production with a real signing key:
+ * DIVINE_RELAY and DEFAULT_RELAYS are module constants, so without an override
+ * every DM path publishes to relay.divine.video and to three public relays
+ * regardless of environment.
+ *
+ * A value that parses to nothing is treated as unset rather than as "no
+ * relays": returning [] from discovery would surface downstream as success===0,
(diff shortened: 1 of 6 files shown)


dcadenas (approved):
DM containment now fails closed without opening real test sockets.

### Changes I pushed

- Prevented containment tests from opening real WebSockets in the Workers pool.
- Treated present `null` and `undefined` relay overrides as contained instead of falling back to production.
- Corrected the description to match the delivered containment behavior.

### Checks

- `npm run lint`
- `npm test`: 84 files, 1,586 tests passed
- Current-head CI: lint, tests, semantic title, and CLA passed
- `repo-validate` pack `35bc323`: path-derived tests passed; five gaps remain because the root and declared CI lint, multi-command test, and reusable semantic workflow are not locally runnable by the validator

PR #200, opened 2026-08-13
fix(reports): stop dropping relay reports on the NIP-89 client tag

realmeylisdev:
## Problem

The report poller required a `client` tag lowercasing to exactly `divine`, and terminally skipped anything else with a 90-day TTL. Two things were being lost, silently at both ends:

- **divine-web reports are 100% dropped.** It tags `divine-web`, which fails `=== 'divine'`. Sampling the 500 most recent kind-1984 events off `relay.divine.video` (the exact relay the poller reads), **144 (28.8%) from 55 distinct reporters** were being discarded — including 8 first-party divine-web reports carrying `NS-underageUser`, `NS-harassment`, and `NS-violence`. `divine-relay-manager`'s `TRUSTED_CLIENTS` allowlist already includes `divine-web`, so the value is intentional and this matcher is the one that drifted.
- **A divine-mobile user who turns off client attribution loses every subsequent report** — divinevideo/divine-mobile#6592. NIP-89 makes the tag optional and explicitly asks clients to let users opt out of it; NIP-56 defines no `client` tag at all. Every report dropped this way was spec-conformant.

Neither end could see it: the user gets the normal confirmation, and the six skip reasons were summed into one `skipped` counter.

## Approach

The tag is unsigned and forgeable from four public string literals in `nip89_client_tag.dart`, so it was never an authorization signal. This demotes it to a trust signal:

- Every valid, video-targeted kind-1984 is **recorded for review**, whatever client sent it.
- An **untrusted-client** report lands on the `relay-report-untrusted` source, so its reporter is excluded from `escalationReporterCount` and cannot supply half the 2-reporter `AGE_RESTRICTED` threshold a later authenticated report would cross. This reuses the boundary that already keeps report DMs review-only rather than inventing a second one.
- `RELAY_REPORTS_REQUIRE_DIVINE_CLIENT` → `TRUSTED_REPORT_CLIENTS`, a configurable case-insensitive allowlist defaulting to relay-manager's `diVine,divine-web,divine-mobile`.
- The single `skipped` counter splits into a per-status breakdown, persisted to KV instead of discarded.

**No report gains automatic power.** Relay reports already never auto-escalated on their own (`allowAutoAgeRestrict` defaults to `source === 'user-report'`); the only thing the client tag ever changed was escalation-counting, which this preserves.

## Blast radius

This changes production report ingestion the moment it merges — deploy is automatic on push to `main`.

Volume: E1 implies roughly **+40%** into the review queue (144 newly recorded against 356 that pass today). The untrusted source is recorded rather than dropped, so it is available to query in D1 on `user_reports.source` — but that is a SQL-level fact only. There is **no** queue-level filter or sort for it today, and the swipe queue does not even display it: `/admin/api/videos` selects `ADMIN_VIDEO_COLUMNS` off `moderation_results` and filters on `action` / `moderated_at`, and that column list excludes `raw_response`. A "hide untrusted by default" switch is a follow-up, not something this PR delivers. **This is the trust-and-safety capacity call and is yours to make** — happy to hold the merge until you've sized it.

Not addressed here: the ~144 reports already tombstoned in KV under the old 90-day skip are still readable and could be re-driven. Leaving them is a choice; say the word and I'll write the backfill.

## Testing

- `npm test` — 1593 passed / 84 files
- `npm run lint` — clean
- New regression tests: a report with **no** client tag reaches `recordReportForReview` (fails on today's `main`); `divine-web` records as trusted; a third-party client records as untrusted; `relay-report-untrusted` counts for display but not escalation, upgrades to escalation-eligible when the same reporter later reports through a trusted path, and never downgrades one that already was.

Refs divinevideo/divine-mobile#6592

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Because the service is report-driven, the queue consumer for the legacy `video-m
 ## Features
 
 - **Human review dashboard** — swipe-review queue, per-category verification (confirm/reject an individual detection), direct-message templates for creator communication, stats, and tunable classifier thresholds. Served behind Cloudflare Access on `moderation.admin.divine.video`.
-- **Report ingestion** — authenticated client reports, inbound public NIP-56 (kind `1984`) reports polled from the relay, and NIP-17 report DMs to the moderation@ inbox. Both Nostr paths are review-only and never auto-escalate, since client tags and reporter pubkeys are self-asserted public signals; report DMs additionally do not count toward the distinct-reporter threshold that the authenticated path escalates on. Only `POST /api/v1/report` can produce an automatic `AGE_RESTRICTED`.
+- **Report ingestion** — authenticated client reports, inbound public NIP-56 (kind `1984`) reports polled from the relay, and NIP-17 report DMs to the moderation@ inbox. Both Nostr paths are review-only and never auto-escalate, since client tags and reporter pubkeys are self-asserted public signals. A relay report is ingested whatever client it came from; the NIP-89 `client` tag only decides whether its reporter counts toward the distinct-reporter threshold the authenticated path escalates on — reports from clients outside `TRUSTED_REPORT_CLIENTS`, like report DMs, do not. Only `POST /api/v1/report` can produce an automatic `AGE_RESTRICTED`.
 - **Content categories** — nudity/NSFW, violence, gore, weapons, recreational drugs, self-harm, hate speech / offensive symbols, AI-generated, and deepfake, plus content-warning labels for alcohol, tobacco, medical, and gambling.
 - **AI-generation & provenance signals** (surfaced to moderators as review context, not auto-enforcement):
   - `divine-ai-detector` for per-signal detection, with vendor fallback to Hive and Reality Defender.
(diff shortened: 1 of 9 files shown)


mbradley (approved):
Thanks for catching this. The drift in how Divine clients identified themselves has been chased down in other corners too. It is frustrating. The old `client === 'divine'` check was throwing away about 29% of reports, and most of that was our own traffic: every divine-web report, plus mobile reports from anyone who turned off client attribution. Recording all of them and firewalling the unverifiable ones out of auto-escalation (the new `relay-report-untrusted` source) is the correct shape, and moving the escalation gate to `escalationReporterCount` actually tightens the age-restrict path rather than loosening it. Tests are solid and the trust boundary is provable. Nice work, Meylis, especially calling the capacity question out yourself instead of burying it.

On capacity: land it. I think we can take the ~40% more review volume, and we will have more pressure anyway. I'm not sure that "we can't take it" is a reason not to get legit Divine user reports to us. The real scaling pressure is still ahead, and that's what we're getting Osprey ready for, so low-stakes moderation can be handled by threshold rules instead of the queue. This PR actually helps that: it starts building the graded report signal Osprey will reason over, instead of dropping it on the floor.

Two small things below, both non-blocking.

**1. Stale comment, `src/reports.mjs:40-47`.** The `// The one report source...` block still says "one," but there are two non-escalating sources now, and the block right under `NON_ESCALATING_SOURCES` already explains both. It is stale and orphaned (it floats above `findReportByReporter`, which is unrelated). Suggest just deleting it. (Left it here rather than inline since those lines aren't in the diff.)

**2. The "filterable" claim.** On "the untrusted source is queryable, so the queue can be sorted or filtered by it": did you mean queryable in the DB, or filterable in the review UI? As far as I can tell the swipe queue filters on `moderation_results.action`, and the trusted/untrusted signal only lives on `user_reports.source` and in the `raw_response` JSON, so there's no queue-level filter today. Not asking you to build one for this PR, just making sure we're not banking on a "hide untrusted by default" switch that would actually need a follow-up.

PR #204, opened 2026-08-21
fix(dm-reader): handle NIP-42 AUTH and stop advancing checkpoint on failed sync

realmeylisdev:
## Summary

Handle **NIP-42 AUTH** in the gift-wrap (kind 1059) DM reader, and stop advancing the inbox checkpoint on an incomplete sync. Part of the rollout for divinevideo/divine-funnelcake#591 (relay-side NIP-42 gate).

## Motivation

The Divine relay is adding NIP-42 AUTH that gates kind-1059 reads to the p-tagged recipient (funnelcake#591). Today `dm-reader.mjs`'s `fetchGiftWraps` handles only `EVENT`/`EOSE` — no `AUTH`, no `CLOSED`. Under the gated relay it would ignore the challenge, wait the full 15 s timeout, and resolve an **empty array while logging "0 errors"** — *and* the checkpoint advanced unconditionally, so reported DMs in that window would be **lost permanently** (self-heal only within the ~2-day overlap). The moderator account holds `NOSTR_PRIVATE_KEY`, so it can authenticate.

## What changed (`src/nostr/dm-reader.mjs` + test, +410/−32)

- **AUTH**: on `["AUTH", challenge]`, sign a kind-22242 event (`finalizeEvent(makeAuthEvent(relayUrl, challenge), NOSTR_PRIVATE_KEY)`) and reply `["AUTH", event]`.
- **OK**: match our auth event id; `true` → authenticated; `false` → reject loudly.
- **CLOSED**: `auth-required:`/`restricted:` prefix → expected pre-auth close; any other reason → reject loudly (mirrors `report-poller.mjs`).
- **Re-REQ**: `maybeResendReq()` re-sends the original REQ exactly once, gated on **both** AUTH-OK and the auth-required CLOSED having arrived (order-independent; a no-auth relay sends neither and never re-REQs).
- **Checkpoint guard**: `fetchGiftWraps` now returns `{events, complete}`; `syncInbox` advances `dm-inbox:last-sync` **only when `complete`** (EOSE received, after any auth). Timeout / close-without-EOSE / auth-refused / non-auth CLOSED / WS error leave the checkpoint so the next tick retries the same window and self-heals.

Verified against the relay side (funnelcake#591): matching `auth-required:`/`restricted:` prefixes, lenient `relay`-tag matching, and same-subid re-REQ served post-auth.

## Test / validation

- `npx vitest run src/nostr/dm-reader.test.mjs` → **33** (26 existing + 7 new: gated AUTH+CLOSED→re-REQ→events; proactive AUTH→EOSE; loud reject on non-auth CLOSED; loud reject on OK-false; checkpoint advances on EOSE / held on close-without-EOSE / held on auth-refused)
- `npx vitest run` (full suite) → **1600/1600**
- `npm run lint` → clean

**Inert until the relay enables the gate.** Safe to merge in any order relative to the relay change.

## ⚠️ One behavior change to note

`/admin/api/messages/sync` (`index.mjs:3764`, not wrapped in try/catch) now returns **500 on a hard failure** (non-auth CLOSED / OK-false / WS error) instead of a silent `{synced:0}` 200 — deliberate "fail loudly" behavior, consistent with the pre-existing WS-error reject. A **timeout still resolves** (now `complete:false`) so it stays 200, and the **cron path already catches + logs**. If you'd rather the admin endpoint translate the reject into a structured error response, that's a small follow-up; left as-is to keep scope on `dm-reader.mjs`.

### DIFF
--- a/src/nostr/dm-reader.mjs
+++ b/src/nostr/dm-reader.mjs
@@ -394,11 +394,9 @@ export function fetchGiftWraps(relayUrl, filter, env) {
     const reqMessage = JSON.stringify(['REQ', subscriptionId, filter]);
 
     const timeout = setTimeout(() => {
-      try {
-        if (ws) ws.close();
-      } catch {}
       // No EOSE means the read was cut short. Return what arrived but mark it
-      // incomplete so the caller leaves the checkpoint where it was.
+      // incomplete so the caller leaves the checkpoint where it was. finish()
+      // closes the socket.
       console.warn(`[DM-READER] WebSocket timeout, returning ${events.length} events collected so far (incomplete)`);
       finish({ events, complete: false });
     }, 15000); // 15 second timeout for potentially many events
@@ -407,6 +405,13 @@ export function fetchGiftWraps(relayUrl, filter, env) {
       if (settled) return;
       settled = true;
       clearTimeout(timeout);
+      // Close the socket on every exit. The reject paths (auth refused, non-auth
+      // CLOSED, AUTH-sign / WS / setup error) previously returned without closing,
+      // leaking the connection until isolate teardown. close() is idempotent, so
+      // the EOSE and timeout paths are unaffected.
+      try {
+        if (ws) ws.close();
+      } catch {}
       if (resultOrError instanceof Error) {
         reject(resultOrError);
         return;
@@ -456,9 +461,9 @@ export function fetchGiftWraps(relayUrl, filter, env) {
         }
 
         if (data[0] === 'EOSE' && data[1] === subscriptionId) {
+          // Polite NIP-01 close of our subscription; finish() closes the socket.
           try {
             ws.send(JSON.stringify(['CLOSE', subscriptionId]));
-            ws.close();
           } catch {}
           finish({ events, complete: true });
           return;
--- a/src/nostr/dm-reader.mjs
+++ b/src/nostr/dm-reader.mjs
@@ -4,9 +4,10 @@
 // ABOUTME: NIP-17 DM inbox reader for moderation conversations
 // ABOUTME: Syncs gift-wrapped DMs from relay and stores in D1 dm_log
 
-import { getPublicKey } from 'nostr-tools/pure';
+import { getPublicKey, finalizeEvent } from 'nostr-tools/pure';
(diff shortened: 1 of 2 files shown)


mbradley (approved):
Approving. This closes a real gap: once the relay starts requiring authentication to read gift-wrapped DMs, the old reader would have quietly collected nothing and then moved its checkpoint past those messages, so reported DMs in that window would have been lost for good. Holding the checkpoint until the relay actually finishes the read is the right fix, and the tests cover the important cases.

@realmeylisdev I pushed three small commits on top (`8cba76d`, `2ab5ccb`, `5af4050`) — please look them over before you merge; they're independent, so drop any you'd rather not take.

**1. Close the connection on the failure paths** (`8cba76d`). When the reader gives up (authentication refused, an unexpected close, a connection error), it was returning without closing the socket — only the success and timeout paths closed it. Moved the close into the one shared exit point so every path cleans up, matching what the report poller already does.

**2. Guard one comparison** (`2ab5ccb`). The check that matches the relay's answer to our authentication attempt could accidentally match a malformed message that arrived before we'd sent anything, because both sides were empty at that point. Added an explicit check so that can't happen. A well-behaved relay never sends that, so this is just belt-and-suspenders.

**3. Handle the "not allowed" close as final** (`5af4050`). The code treated two different relay responses the same way: one means "authenticate and try again", the other means "you authenticated but you're not allowed". The second can't be fixed by retrying, so retrying just draws the same refusal. Made it stop and report the error on the first response instead, which also matches what the protocol says these two responses mean. This was never a silent failure — the relay sends its authentication challenge as soon as you connect, so the reader always authenticates first — but reporting it right away is cleaner and saves a pointless retry. Added a test.

**One thing worth a sanity check:** during a prolonged authentication outage the checkpoint intentionally stops moving and every run re-reads the same window. That's the correct trade (far better than losing messages), but the read is capped at 200 events, so if a very large backlog built up while authentication was broken, the oldest messages past that cap wouldn't be picked up until things recover. Nothing to fix here — just flagging that the fix trades "lose messages permanently" for "retry forever, 200 at a time".

Full test suite green locally (1601 passing).

